PR #308, opened 2014-04-24
Statistics db

Thorsten-Sick:
Adding a task statistics db and a web.py extension to view it. Goal of the db is to profile the tasks for performance, success and errors. I hope to enable me and others to identify issues (for example with specific vms) and performance bottlenecks


jbremer:
I like the idea but it's been a while and thus the code does not merge cleanly and will require quite some merging. Also it's a bit bloat here and there, but perhaps something similar will be created in the distant future.

PR #327, opened 2014-05-12
Profiling log

Thorsten-Sick:
Adds more DEBUG log lines to the Cuckoo output (activate with cuckoo --debug). The log can be processed with "utils/stats.py --profile " and show how much time the process spent in where.


jbremer:
Would be more appropriate to introduce a separate table for this in the database with one timestamp per profiling entry. Could be nice..

PR #362, opened 2014-07-31
Bugfix enhanced behavior

Thorsten-Sick:
Enhanced behavior had some issues based on a process spawning several childs and those use the same handle-ids for different objects. This should fix them


jekil:
Do you have a test case we can use to reproduce the bug?

jbremer:
With all of the work on the new monitor this work has, unfortunately, been superseded. (As you know all of the registry is now done natively in the monitor, rather than on the processing side).

PR #381, opened 2014-08-25
Cache for behavior processing. Large improvements in processing speed

Thorsten-Sick:
The API data is processed only once. The results are taken from the memory cache. Processing of one of my insane analyses went from 30 minutes to 11. Adding ~ 600MB to RAM usage. (the json report is 300MB, so this is a very special case). And it scales better with a growing number of signatures or reporting modules. Normal test cases show similar results.


jbremer:
All of this is being reworked for the upcoming `2.0` release. Will have to do some new performance analysis then..

PR #397, opened 2014-10-19
Modified processing results from dict to defaultdict

knqyf263:
Failed to generate HTML report with 'dict object' has no attribute 'static'.
This is why ReportHTML try to access to the key that dictionary does not have.


jbremer:
I don't think this a proper solution for the problem. Mostly because some entries in the `results` dictionary are lists and other are dictionaries. E.g., the `static` entry is a `dict`, but the `dropped` entry is a `list`.
If this bug is still present, then we have to open an issue to get it resolved specifically for the `static` entry.

PR #399, opened 2014-10-19
Checking for keys in processing results dict

knqyf263:
Failed to run the reporting module "MongoDB"
because this module does not check for keys in dict.


jekil:
Thank you so much for your contribution, it was not merged because it there are some pieces of related fixes in other commits, although your contribution help us to spot a couple of other bugs. Thanks

PR #411, opened 2014-11-17
Added creation/deletion of registry keys to static HTML report summary

depaolism:
These changes allow for more detailed information about the registry to be displayed in the web gui. It does not differentiate between key addition and modification and so treats all writes the same. Writes have green text and are prefixed with 'w+', deletions have red text and are prefixed with '-'. An example is below:

stry

### DIFF
--- a/data/html/sections/behavior.html
+++ b/data/html/sections/behavior.html
@@ -45,10 +45,20 @@
         </div>
         <div class="well well-small">
             <b>Registry Keys</b>
-            {% if results.behavior.summary["keys"] %}
-                <ul>
-                    {% for key in results.behavior.summary["keys"] %}
-                    <li><span class="mono">{{key}}</span></li>
+            {% if results.behavior.enhanced.registry_events %}
+                <ul style="list-style: none;">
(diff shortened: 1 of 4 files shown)


depaolism:
Just checking on the status of this pull request. Is there something that I could do the help clean up behavior.py?

gtback:
Is there a chance this will make it into the 1.2 release?

botherder:
I need to work on it.

Firstly, whenever the results structure changes, it causes compatibility issues. Secondly, the HTML report is pretty much abandoned, so this should be ported to the Django interface.
Lastly, as it is replacing the Summary, at that point I'd rather replace it completely for the other sections as well.

depaolism:
Screenshot of changes made by most recent commit.
files_summary

jbremer:
This functionality has been integrated as part of the `monitor` branch. Both the `json` report and the (deprecated) `HTML` report now show a summary with all these things in there, sorted by the type of event (`file written`, `regkey created`, etc). Integration for Django is still pending, but this is not the goal of this pull request anyway. So, closing this PR.

Issue #419, opened 2014-12-03
Add url callback in the Cuckoo REST API (/tasks/create/file and /tasks/create/url)


henriquemenezes:
Add a field callback to resources (/tasks/create/file and /tasks/create/url) on REST API. 
This field can be used to report the a finished analysis.

jekil:
Closed as duplicates #420

PR #420, opened 2014-12-04
Added callback url to REST API.

henriquemenezes:
Issue #419 resolved. 
Recently I created an issue to add a url callback to be called after analysis finished:
`(Add url callback in the Cuckoo REST API (/tasks/create/file and /tasks/create/url) #419)`
I believe this feature is interesting to report a finished analysis.


jbremer:
Good idea and has been requested more often. However, should be configurable from the configuration files as well (while still being able to override it using a per-task argument).
And also, I know we still use it ourselves as well, but *one does not simply use `urllib`*.

jbremer:
And this will require a database migration, but that's not an issue for you.

jbremer:
I'm going to close this. This would be much more effective as a 5-line reporting module, aside from the fact that this is a very specific use-case for which most users will want a small change to their likings.
Thanks for the effort, though!

PR #424, opened 2014-12-11
Better Signatures

gaelmuller:
Adds two helpers in signatures: `Signature.add_match()` and `Signature.has_matches()` to simplify addition of matching data to the signature (no need to update `Signature.data` manually anymore). It is now easy to add one (or several !) match(es) to the Signature.

Also adds support for better display of matching data in the web interface. Simple example, displaying a dict:

image

Support for special type `'api'` to use with evented signatures, in order to extract the matching API calls. Example:

image

The arrow on the right allows the user to jump directly at this API call in the Behavioral tab.

With this functionality, it was also obvious to implement a simple search in the Behavioral logs:

image

For this change to really mean something, I also updated every signature in the community repository to use this feature. This way, you can click on every signature with a high level of confidence that it will display the matching data. See PR https://github.com/cuckoobox/community/pull/46

### DIFF
--- a/docs/book/src/customization/signatures.rst
+++ b/docs/book/src/customization/signatures.rst
@@ -261,6 +261,65 @@ You can find many more example of both evented and traditional signatures in our
 
 .. _`community repository`: https://github.com/cuckoobox/community
 
+Matches
+=======
+
+Starting from version 1.2, signatures are able to log exactly what triggered
+the signature. This allows users to better understand why this signature is
+present in the log, and to be able to better focus malware analysis.
+
+Two helpers have been included in order to specify matching data.
+
+.. function:: Signature.add_match(process, type, match)
+
+    Adds a match to the signature. Can be called several times for the same signature.
+
+    :param process: process dictionary (same as the ``on_call`` argument). Should be ``None`` except when used in evented signatures.
+    :type process: dict
+    :param type: nature of the matching data. Can be anything (ex: ``'file'``, ``'registry'``, etc.). If match is composed of api calls (when used in evented signatures), should be ``'api'``.
+    :type type: string
+    :param match: matching data. Can be a single element or a list of elements. An element can be a string, a dict or an API call (when used in evented signatures).
+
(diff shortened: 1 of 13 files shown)


bit111:
Hi @gaelmuller,
I created a new signature (and I hope it will be interesting for Cuckoo community) but I have not well understand how to format output.
I read this topic about introduction of _Signature.add_match()_ but I have not understand how to use it.
I wish to realize a signature output similar to your first example _"The binary likely contains encrypted or compressed data."_
I tried to use 
```
formatted_string = "%s \n %s \n %s " % (field1, field2, field3)
self.mark_ioc("http", formatted_string)
```
but fields are not showed each one in a new row.
I also tried to use 
```
self.add_match(None, 'url', 'http://site1.com')
self.add_match(None, 'url', 'http://site2.com')
self.add_match(None, 'url', 'http://site3.com')

```
but I have an error during signature execution.

gaelmuller:
Hello @bit111,

API has changed since this PR. From what I understand, `add_match` has been replaced by `mark_call`, `mark_ioc`, `mark_vol` and `mark`.

Web pages do not understand "\n" has a newline, which is why they are not displayed. Depending on your case, you might want to try:

```
self.mark_ioc("http", field1)
self.mark_ioc("http", field2)
self.mark_ioc("http", field2)
``` 
or
```
self.mark(field1=field1, field2=field2, field3=field3)
```

AFAIK, it is not possible anymore to write a signature that will have a similar output to my example.

bit111:
Hi,
in my signature I display some stolen data from user. For this reason I would like to well format output in way to show this:
_____________________________
_This signature appears because some data were stolen. These are the stolen data:_
**Field 1 Title** ---------------- "content"
**Field 2 Title** ---------------- "content"
**Field 3 Title** ---------------- "content"
_____________________________
This block (limited by underscore) is printed in a loop: for this reason I am looking for a way to print more block of this (if there are more matches) in the same signature.
Do you think it is possible?

PR #434, opened 2015-01-08
XenServer Machinery

ameily:
This merge request implements the XenServer machinery. The machinery communicates to a XenServer instance through the XenAPI library, which uses an HTTP(S) XML RPC interface. The machinery supports the following functions:

* Starting and stopping VMs: `start()` and `stop()`
* Checking the power state of a VM: `_status()`
* Listing available VMs: `list()`
* Reverting a VM to a previous state

The XenServer machinery supports two methods of VM restoration:

* Memory and disk snapshots
* Disk resetting on boot

Memory and disk snapshots require that the VM has the Xen guest tools installed, which may not be possible in some environments. Therefore, the machinery also supports launching VMs that have their disks reset on boot, which is configured on a per VM basis in the XenServer instance.

The documentation has also been updated to include XenServer related configuration and usage.


SnakeByteIL:
Great work! specially since newer versions of esxi (free) don't have API.

botherder:
It looks good to me, but it would be good to have some proper testing. For significant contributions, please also remember to check out this http://cuckoofoundation.org/contribute.

ameily:
Signing the agreement shouldn't be problem. I'll take a look at it tomorrow.

As far as testing, I tested the machinery against a XenServer instance that I already had running with multiple VMs and different configurations. Beyond that, are you looking for anything formal?

botherder:
No, not formal, but it's generally good to have feedback from other users so that they can verify the setup and configuration procedures as well as perhaps reveal issues that only different installations might raise.

wealthychef:
ameily:
I spun this up on my i7 box. Here's my notes:

o The documentation doesn't mention installing xenapi. Gave me an error - 'pip install xenapi' fixed it
o The configuration section was confusing. I entered machine=xen and wondered why it was failing (machine=xenserver for those wondering)
o When choosing to use VM disk resetting (Boot from disk section), it should mention that the VM must be shut down before executing the commands on XenServer
o Maybe it was the machine I had running XenServer, but suspending, resuming, restarting, and ejecting disks from the vm through XenCenter was sluggish (~1 minute for each command) 
o There were times when the cuckoo job would hang and cuckoo.py had to be restarted to resume. Probably a cuckoo thing and not a Xen thing

ameily:
Thanks for the feedback @wealthychef! I've clarified spots of my documentation that you pointed out. Specifically, I modifed the `Installation Preparing the Host -> Requirements` and `Preparing the Guest -> Saving the Virtual Machine -> Booting from Disk` sections. Could you please review it and make sure it good to go?

Also, I'd like to confirm that the hanging issue you are seeing is not the XenServer machinery. I tested the machinery locally again and found that the XenAPI is not thread safe. I've fixed this limitation with my recent commits. My guess is that you were either seeing this issue or you did not specify a runtime timeout for the samples. Can you please retest and verify that the issue is not with the XenServer machinery?

wealthychef:
I checked out the documentation and it looks like the ambiguities were addressed. I also re-ran my previous tests to attempt to get cuckoo to 'hang', but it ran smoothly through everything I threw at it. Thanks for the quick response!

botherder:
Unless @jekil has some issues with it, I'm ok for merging.

jekil:
@botherder ok for me

jekil:
@ameily thanks a lot for this contribution, it is really high-quality code.

alexandrafengfei:
Hi. Now I am using Ubuntu with Xen hypervisor. Could I use the XenServer machinery? Or how could II modify it? The xen hypervisor does not have a XenServer URL.

ameily:
The stock Xen hypervisor is not compatible with the XenServer machinery. However, you should be able to use the libvirt machinery.

alexandrafengfei:
Thank you! Just to be sure, I should modify the machinery=xen in the cuckoo.conf, right?

PR #451, opened 2015-01-29
Update packages.py

SecTecRes:
Support for Excel files detected as "Microsoft Excel" (not only "Microsoft Office Excel")


jekil:
Thanks for the contribution.

PR #452, opened 2015-01-29
Running generic files

SecTecRes:
Running 'cmd /c start "calc.exe"' just opens a Command Line with the title of 'calc.exe' instead of running 'calc.exe' ("calc.exe" is handled as the title and not the command/program.) we just need to define a title to fix it.
the syntax of 'start' command in windows is:
START ["title"] [/D path] [/I] [/MIN] [/MAX] [/SEPARATE | /SHARED]
      [/LOW | /NORMAL | /HIGH | /REALTIME | /ABOVENORMAL | /BELOWNORMAL]
      [/NODE <NUMA node>] [/AFFINITY <hex affinity mask>] [/WAIT] [/B]
      [command/program] [parameters]

### DIFF
--- a/analyzer/windows/modules/packages/generic.py
+++ b/analyzer/windows/modules/packages/generic.py
@@ -12,5 +12,5 @@ class Generic(Package):
 
     def start(self, path):
         cmd_path = self.get_path("cmd.exe")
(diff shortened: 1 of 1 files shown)


jekil:
Thanks s much.

Issue #453, opened 2015-02-01
cuckoo.py --clean Database() logging handler doesn't work


KillerInstinct:
When we run --clean we are unable to catch errors due to not initializing the log handler (since we are deleting the logs) The hacky fix is to try to clean the database first with a try/catch to log errors before we delete the logs. This way we can at least output to stdout.

KillerInstinct:
Below is a better representation of my hacky fix above. :)
```python
def cuckoo_clean():
    try:
        init_logging()
    except IOError:
        # We tried to clean when the log file doesn't exist. If we want
        # logging for the Database class log hander, we need to reinitialize
        # the log file.
        create_structure()
        init_logging()

    # Initialize the database connection.
    db = Database()
    print db.count_tasks()
    # Drop all tasks.
    db.drop_tasks()
    print db.count_tasks()
    # Drop all samples.
    db.drop_samples()
    paths = [
        os.path.join(CUCKOO_ROOT, "db"),
        os.path.join(CUCKOO_ROOT, "log"),
        os.path.join(CUCKOO_ROOT, "storage"),
    ]

    # Delete various directories.
    for path in paths:
        if os.path.isdir(path):
            try:
                shutil.rmtree(path)
            except (IOError, OSError) as e:
                log.warning("Error removing directory %s: %s", path, e)

    # Delete all compiled Python objects ("*.pyc".)
    for dirpath, dirnames, filenames in os.walk(CUCKOO_ROOT):
        for fname in filenames:
            if not fname.endswith(".pyc"):
                continue

            path = os.path.join(CUCKOO_ROOT, dirpath, fname)

            try:
                os.unlink(path)
            except (IOError, OSError) as e:
                log.warning("Error removing file %s: %s", path, e)
```

jekil:
Uhm, this is not working, because you are logging (for example in the last line) to a deleted file anyway.
I got: 
IOError: [Errno 2] No such file or directory

PR #455, opened 2015-02-01
Fix up --clean for MySQL

KillerInstinct:
Added a function to drop guests as there was an issue in MySQL where we couldn't delete the samples table due to a FK constraint with tasks, which had a FK contraint to the guest table. We also drop the machines-related tables; should probably drop errors and tags tables in a future commit to be truly clean. We were also missing commits to the drop statements for the samples and tasks tables. These errors were invisible due to the missing log handler, which I have logged under issue #453.

### DIFF
--- a/cuckoo.py
+++ b/cuckoo.py
@@ -100,6 +100,12 @@ def cuckoo_clean():
     # Initialize the database connection.
     db = Database()
 
+    # Drop machines and machines_tags.
+    db.clean_machines()
+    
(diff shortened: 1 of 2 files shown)


jekil:
Thank you very much for your contribution, although I tried a more sqlalchemyish fix (db46352). Let me now if it's ok with mysql (postgres is fine).

KillerInstinct:
Works in my env with MySQL -- thanks!

PR #457, opened 2015-02-02
Add MSI Analyzer Package

KillerInstinct:
Create an MSI analyzer package using msiexec.exe. Add a detection to automatically use the package for files submitted with a .msi extension. (File magic is unreliable)


jekil:
Thanks a lot, nice contribution.

PR #461, opened 2015-02-03
Added python package support for analyzer

XayOn:


jekil:
Thanks for your contribution, but we already have a python analysis package (it's called "python.py") and your analysis package is implemented in the "old way", i mean directly dealing with process.

XayOn:
Oh, sorry we were a bit outdated on that.
I've seen that you already have added all the pythonpaths we had on this pull request to your analysis package. 

Thanks.

PR #464, opened 2015-02-04
processing-memory: workaround for bug in volatilty when using libvirt/qe...

mak:
there is a problem with volatility when using qemu core dumps - and its marked wont-fix (issue nr 10)
this a workaround as proposed in issue - use psscan to get system dtb and apply it


jekil:
Thanks for contributing

jekil:
Great patch, thanks a lot!

PR #468, opened 2015-02-06
Allow --clean to drop MongoDB database

KillerInstinct:
Load the reporting config and drop the MongoDB database if we're configured to use it.

### DIFF
--- a/cuckoo.py
+++ b/cuckoo.py
@@ -11,6 +11,7 @@ import sys
 
 try:
     from lib.cuckoo.common.logo import logo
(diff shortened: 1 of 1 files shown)


KillerInstinct:
Feel free to rearrange the import to the top. I normally imported after the checks so it's not loaded if someone has that reporting module turned off, but since this is the --clean and we should be exiting after we deleted everything, it could probably be included at the top with the other imports.

jekil:
Trust, i was just thinking about this feature just 2 days ago! It's great to have it already implemented, thanks a lot for your contribution.

PR #480, opened 2015-03-02
Added possibility to specify Volatility profile per machine

SteveClement:
Added a profile option in the config file (virtualbox.conf) to specify a volatility profile, bypassing auto-detection and speeding up the analysis of the submissions.

### DIFF
--- a/modules/processing/memory.py
+++ b/modules/processing/memory.py
@@ -8,6 +8,7 @@ import logging
 from lib.cuckoo.common.abstracts import Processing
 from lib.cuckoo.common.config import Config
 from lib.cuckoo.common.constants import CUCKOO_ROOT
+from lib.cuckoo.core.database import Database
 
 try:
     import volatility.conf as conf
@@ -916,6 +917,8 @@ class VolatilityManager(object):
         self.taint_pid = set()
         self.memfile = memfile
 
+        db = Database()
+
         conf_path = os.path.join(CUCKOO_ROOT, "conf", "memory.conf")
         if not os.path.exists(conf_path):
             log.error("Configuration file volatility.conf not found".format(conf_path))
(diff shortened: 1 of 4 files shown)


KillerInstinct:
I took a stab at this a while back also, feel free to use/refence:
https://github.com/KillerInstinct/cuckoo-modified/commit/0b925f567e05535f6ce9c5ecc1d1e211f9c8d39d

botherder:
Your implementation looks good. I also don't see the need to store the information in the DB.

jbremer:
I'd be in favor of propagating this information through the `options` field. This should allow to replace the somewhat ugly `profile = Config(manager).get(vm).get("mem_profile")` by something such as `self.task.options["volprofile"]`. Thoughts?

jbremer:
Given the *size* of this feature - considering volatility is a pretty addition to Cuckoo, I'd propose to add a column to the `machines` table to reflect the volatility profile. That'd make @KillerInstinct's patch a bit more properly developed.

modules/processing/memory.py
```diff
@@ -929,9 +932,25 @@ def __init__(self, memfile, osprofile=None):
             if pid:
                 self.mask_pid.append(int(pid))
 
+        response = {}
+
+        task = db.view_task(db.list_tasks(limit=1)[0].id, details=True)
```
jbremer: What is happening here; getting an instance of the current task? If so we can just replace this with `self.task`.

jbremer:
I'll put this on my personal todo list for the upcoming months (as I'll also be using this functionality), but for reasons stated above I won't be merging this PR. So will close it until I find the time to properly implement it, perhaps also based on @KillerInstinct's commit(s). Even though it's been almost 2 years now, thanks for your contributions ;-)

SparkyNZL:
that would be awesome, I have mine default to Win7 but its not that good
for WinXP or Win10 :(

jbremer:
For what its worth, support for per-VM `osprofile`'s has been released as part of version `2.0.2`.

PR #486, opened 2015-03-03
Add PEiD Signatures to Django

KillerInstinct:
They are already collected by the static processing module.


jekil:
Thanks for contributing.

PR #487, opened 2015-03-04
Add an open button and use a button blacklist.

KillerInstinct:
This fixes a bug/race condition for when we have a window with two buttons: Run and Don't Run. Cuckoo would sometimes end up clicking Don't Run as "run" was in the button text. The introduction of a black list allows us to be more explicit. Also add the 'open' button for windows where our only choices are 'open', 'open folder', or 'cancel'. (In the case of running JavaScript files with IE)


jekil:
Thanks for the contribution

PR #533, opened 2015-04-24
Translate of readme for the Portuguese

igorcaavalcante:


jbremer:
I'm not sure if we want to pursue this road. If there's a README in Portuguese, then one might assume the same goes for the entire documentation. It's not like we update the documentation on a daily basis, but relying on external contributors to keep their language documentation (or readme) up-to-date may not work in the long run.

In any case, if we stick to just a translated README, then I'd still suggest to rename those to e.g. `README.md.br` or `README.md.pt` or whatever extension you deem appropriate.

jbremer:
Assuming you won't be doing a full documentation translation, I guess there won't be a point in having a Portuguese readme. Anyone have an opinion on this? Otherwise this will be closed in a few weeks..

jhg:
I don't understand this, then all coments on code will be translate to all languages? Not is good this, I think. Is good for example documentation (web, PDF, etc) translate and maintain this, but readme is like comments in code, is hard use all languages in a project.

jhg:
And travis fail for `ABAIXO HÁ UMA TRADUÇÃO PARA O PORTUGUÊS` I think that travis has problems with all chars of unicode, or other about charset.

igorcaavalcante:
I did it because sometimes i'm too lazy to translate something and i'm sure
other Brazilian also has. But don't worry about it, i understand your point
and I agree this would imply translate all documentation.

Thanks for you time.
Igor

2015-06-04 11:04 GMT-03:00 Jesús Hernández Gormaz <notifications@github.com>
:

jhg:
@igorcaavalcante I am from Spain, I speak spanish, I am also too lazy to translate something and other people from spain also has. And it's same with french, japan, chinese, etc. But not is possible has a readme and comments for each language and all updates (I don't know how use translations for translate the code). I understand you, but we are developers, we do the hard work for the user has a friendy user interface and a soft work. :D

I want proposal that Cuckoo documentation http://docs.cuckoosandbox.org/en/latest/ (https://github.com/cuckoobox/cuckoo/blob/master/docs/book/src/index.rst) will have translation and I can translate it to spanish, @igorcaavalcante do you can translate it to brazilian? @jbremer all right for translate doc to other languages? And the documentation can have the same and more information that readme.

botherder:
If there is an interest in translating the documentation, we can start a transifex and do it properly. Translating the README isn't really useful.

igorcaavalcante:
Thinking good i realize that is not too important for the project, it's
better forgot this. Besides i have to much job right now and i don't have
time to do this in the moment (when i commited i wasn't working).

I appreciate everyone's attention and I hope to contribute to the project
another time and in another way.

Good work, boys :D

2015-06-05 20:36 GMT-03:00 Nex <notifications@github.com>:

PR #536, opened 2015-04-29
Fix a bug where we couldn't delete URL tasks

KillerInstinct:
URL's have no sample and thus do not have a file_id in the target dict.


jbremer:
Haven't tested but looks fine to me. Feedback anyone?

brad-accuvant:
It works, pull it

PR #543, opened 2015-05-10
as of pymongo 3.0, Database.disconnect() as been removed as a synonym for Database.close().

### LINKED ISSUE #522: MongoDB "disconnect" error
Running
Cuckoo v 1.2
PyMongo 3.0

Cuckoo throws an error when trying to close a mongodb connection.
```
2015-04-13 14:21:17,556 [lib.cuckoo.core.plugins] ERROR: Failed to run the reporting module "MongoDB":
  File "/mnt/data/cuckoo/lib/cuckoo/core/plugins.py", line 505, in process
    current.run(self.results)
  File "/mnt/data/cuckoo/modules/reporting/mongodb.py", line 216, in run
    self.conn.disconnect()
  File "/usr/local/lib/python2.7/dist-packages/pymongo/database.py", line 1050, in __call__
    self.__name, self.__client.__class__.__name__))
TypeError: 'Database' object is not callable. If you meant to call the 'disconnect' method on a 'MongoClient' object it is failing because no such method e
xists.
```
Changing `self.conn.disconnect()` to `self.conn.close()` in mongodb.py seems to solve it.

johnfoo:
fixes #522

### DIFF
--- a/modules/reporting/mongodb.py
+++ b/modules/reporting/mongodb.py
@@ -213,4 +213,4 @@ class MongoDB(Report):
 
         # Store the report and retrieve its object id.
         self.db.analysis.save(report)
-        self.conn.disconnect()
+        self.conn.close()


jbremer:
Do older versions of `pymongo` ship with the `close` method? Otherwise something like the following would be more appropriate:

```python
    if hasattr(self.conn, "disconnect"):
        self.conn.disconnect()
    else:
        self.conn.close()
```

johnfoo:
see my comment in the issue thread. but the behavior is not modified because MongoClient.close() was a simple alias to self.disconnect, and commit mongo/mongo-python-driver@7f42430978310e8bd9c4ed2995abd7da24c0f52f simply changed the behavior so calls to MongoClient.close() has the actual logic.

jbremer:
Well, I'll leave it up to @jekil and @botherder, but I approve this PR :)

PR #548, opened 2015-05-13
prototype of a hook mechanism to integrate cuckoo task analysis without polling the API

johnfoo:
not a real pull request per se, more of a prototype to gauge external interest.

i'm integrating cuckoo support in an application we're developing, and all our "resolvers" are asynchronous. as i'd like to limit polling, i'm implementing a hook mechanism to let cuckoo tell an external url when an analysis is complete.

i made a very primitive listener that waits for a POST containing a json dict with ```{'task': <id>}``` and gets ```tasks/report/<id>```

the real mechanism will probably be more [resthooks](http://resthooks.org/) like, with a real subscription mechanism for workers.

what would be your requirements for such a mechanism to be implemented in mainline ?

as a side note, the cuckoo code currently mixes calls to urllib2 and requests, i do not know if there is an 'official' position on which library to use, but i'd advise definitively switching to requests which makes for easier and cleaner code IMHO.

### DIFF
--- a/conf/cuckoo.conf
+++ b/conf/cuckoo.conf
@@ -57,6 +57,12 @@ freespace = 64
 # (web.py, api.py, Django web interface).
 tmppath = /tmp
 
+# cuckoo webhook callback when analysis is complete
+webhook = on
+
+# url to POST to
+webhook_url = http://127.0.0.1
+
 [resultserver]
 # The Result Server is used to receive in real time the behavioral logs
 # produced by the analyzer.
--- a/lib/cuckoo/core/scheduler.py
+++ b/lib/cuckoo/core/scheduler.py
@@ -7,6 +7,9 @@ import time
 import shutil
 import logging
 import Queue
+import urllib2
+import json
+
 from threading import Thread, Lock
 
 from lib.cuckoo.common.config import Config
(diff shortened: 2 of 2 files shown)


johnfoo:
disregard that PR, i just saw #420. that'll teach me to read PRs first

jbremer:
As team we discussed this feature sometime ago (like a year or so), I think at least me and @rep were interested in such feature, but I guess that neither of us actually needed it in the end, and thus no such feature was introduced.
Yeah, urllib is still used for no reason, but for new features I'd suggest requests.

You know you can *just* call `/tasks/list` with `status=reported` or something along these lines (refer to the documentation) in our Cuckoo API every now and then to get a list of completed (and reported) tasks? I mean, it's still polling, but generally works fine for me.
Of course this depends on whether you're using Cuckoo for mass analysis (where a minute delay is no big deal) or whether you want instant results for a few tasks, in the latter case I understand your interest in a notification feature (as we called it when we were discussing it).

johnfoo:
polling is not the easiest option for us due to the structure of the service we have, hence my need for cuckoo to spontaneously report. the final workflow will probably be something along the line of "workers subscribe to report rest hook, workers post tasks to be done by cuckoo, cuckoo publish the json report of tasks to subscribed workers directly".

PR #552, opened 2015-05-20
api.py - Unable to modify tar file format using GET variable

Shane-Carr:
The "tar" variable is not retrieved correctly from the GET request.

eg: /tasks/report<id>/dropped?tar=gz


jbremer:
Thanks! It seems there's a weird error with Travis CI, but other than that the PR looks good. Will be merged soon.

PR #553, opened 2015-05-21
Fix #512

### LINKED ISSUE #512: Integrating YaraRules repository
YaraRules runs a comprehensive collection of Yara rules:
http://yararules.com/rules/

At the moment our Yara rules are not really updated and are very limited. It would be a good idea to find ways to integrate their collection in Cuckoo, perhaps through a dedicated utility script or an additional option in `utils/community.py`?

SpoonBoy:
Made a quick way to download http://yararules.com/rules/ in the `init_yara()`
so they will be updated each time we launch cuckoo

### DIFF
--- a/lib/cuckoo/core/startup.py
+++ b/lib/cuckoo/core/startup.py
@@ -12,6 +12,7 @@ import urllib2
 import logging
 import logging.handlers
 import pwd
+import re
 
 import modules.auxiliary
 import modules.processing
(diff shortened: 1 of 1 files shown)


jhg:
I like it, I will try this.

jhg:
Work fine! I like it. Can pull it? @jbremer

botherder:
I will look into it. I'm concerned about the size of the rules collection and duplication.

jhg:
If it's big for run in startup then is possible move this to a utils script. My data/yara size after run this is of 1.2MB.

brad-accuvant:
I'm not sure they would appreciate the full ruleset being downloaded unnecessarily every time Cuckoo starts.  Local caches and header-based modification checks should be used so that the site isn't unnecessarily abused, similar to what we do in our Feed class: https://github.com/brad-accuvant/cuckoo-modified/blob/master/lib/cuckoo/common/abstracts.py#L1334

SpoonBoy:
indeed, that was a "quick way".
@brad-accuvant you have a lot of other features that interest the community. So i could only encourage you to pull them :)

jhg:
Check of headers can be added in this pull request or when it is in master, not is problem. All set of rules is good, give interesting data about a sample. Also is interesting add after a script for re-check and download with cuckoo running, or a periodic task for this.

And well changes for comunity can be pull request and it is appreciate.

jbremer:
I'm afraid this is not going to cut it for now. I'm definitely in favour of adding more Yara rules, but I think we should make a list of Yara rules that we want to have and place those in the `community` repository.
Placing the Yara rules in the community branch also avoids having to re-download them every start of Cuckoo. I think manual upgrading is better in this case. (What if a new Yara rule with a 99% false positive rate is introduced?)
Going to close this issue for said reasons - feel free to re-open or create new issues if a better solution is found :-)

Issue #554, opened 2015-05-21
Analyzer does not find 64 bit executables when python is 32 bit


worsht:
Steps to reproduce:
- Start from a fresh Windows 7 guest OS
- Install 32 bit python and the agent
- Install 64 bit Microsoft Office
- Try to sandbox a document
- The following error is seen:
```
2015-05-21 15:48:52,314 [lib.cuckoo.core.scheduler] ERROR: Analysis failed: The package "modules.packages.doc" start function raised an error: Unable to find any Microsoft Office Word executable.
```

Root cause:
The code in abstracts.py that resolves executable paths assumes that  %ProgramFiles% points to C:\Program Files and %ProgramFilesx86% points to C:\Program Files (x86), whereas in this case %ProgramFiles% actually  points to C:\Program Files (x86). Since Office is 64 bit and is installed in C:\Program Files, it will never be found.

worsht:
This PR: https://github.com/cuckoobox/cuckoo/pull/555 has a patch

jbremer:
Thanks for the bug report :)

jbremer:
Merged your patch, thanks again.

worsht:
Thank you!

jbremer:
Please comment on https://github.com/cuckoobox/cuckoo/pull/555#issuecomment-108667095 here, thanks!

jbremer:
64-bit analysis has been improved quite a bit and this works perfectly fine now. Going to close the issue as the pull request has been integrated a while ago and works smoothly (in combination with some other minor changes).

PR #555, opened 2015-05-21
#554: Analyzer does not find 64 bit executables when python is 32 bit

worsht:
If %ProgramFiles% resolves to C:\Program Files (x86), we should still yield C:\Program Files. Fixes issue: https://github.com/cuckoobox/cuckoo/issues/554

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,9 +24,14 @@ Thumbs.db
 # Ignore development files
 docs/book/src/_build/
 .idea/
+.project
+.pydevproject
 
 # Ignore Django secret_key
 web/web/secret_key.py
 
 # Ignore yara rules
 data/yara/index_*.yar
+
+# Ignore venv
+venv
--- a/analyzer/windows/lib/common/abstracts.py
+++ b/analyzer/windows/lib/common/abstracts.py
@@ -39,7 +39,10 @@ class Package(object):
             if basedir == "SystemRoot":
                 yield os.path.join(os.getenv("SystemRoot"), *path[1:])
             elif basedir == "ProgramFiles":
-                yield os.path.join(os.getenv("ProgramFiles"), *path[1:])
+		program_files = os.getenv("ProgramFiles")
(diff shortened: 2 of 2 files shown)


jbremer:
So the bug report is good, but the patch can be improved in my opinion. Besides the gitignore stuff and missing indentation, you change one `yield` statement into two `yield` statements.
In 32-bit Python on a 64-bit VM this would result in first yielding the `C:\Program Files (x86)` path, followed by yielding the `C:\Program Files` path - is this desired behavior?
Personally I'd probably go for a very basic `.replace(" (x86)", "")` instead of the whole magic number string slicing, what do you think?

worsht:
Thanks! The desired behavior is to return both C:\Program Files\ and c:\Program Files (x86)\ regardless of whether python is 32bit or 64bit. If we removed the trailing x86, 32 bit executables wouldn't be found.  Thoughts?

worsht:
Ah, I see what you mean. Made the change.

jbremer:
Yeah, due to the `ProgramFiles(x86)` thing 32-bit applications would still be found. Thanks, your patch looks good to go to me.

jbremer:
This *bug* also applies to `C:\Windows\system32` which gets translated into `C:\Windows\SysWOW64`.
The new monitor has work in progress support for 64-bit executables, however, for big binaries (e.g., Internet Explorer, Microsoft Office, etc) various improvements have to be done, involving a lot of debugging.. So actually I'm not sure if this pull request is good at the moment.

Do you have any applications which you'd like to analyze in 64-bit mode at the moment that are different in 32-bit mode? Like I said, for Windows applications and common 3rd party applications I don't think it makes sense, *at the moment*.

worsht:
My use case was to see screenshots of the application loading the doc even if the analysis was incomplete, which this PR solves. But you're right - it makes sense to fully support 64bit applications before attempting to launch them.

manzy90:
Sorry i am having this issue,so how to i fix it what do i need to do?

PR #558, opened 2015-05-28
Added: removal of sorted_pcap data when deleting analyses

carthurb:
When deleting analyses from the web interface, sorted pcap data files are left behind in the reporting database.  Update removes these orphans.


jbremer:
@jekil @botherder This one is for you.

botherder:
Yes, makes sense. Thanks.

PR #559, opened 2015-05-29
Tools

depaolism:
Adding ability to submit an analysis tool to run on a sample within the sandbox. The user is able to specify options for the tool and the sample and the results are returned to the host. 

An example of an analysis tool would be a dynamic unpacker or an execution tracer.

See usage/results.rst & usage/submit.rst for more details.


depaolism:
A database migration needs to be done and I'd be happy to do it if you like the changes overall.

init99:
Super interesting. Could this process also be used if the malware requires components (e.g. an EXE along with DLL(s))?

depaolism:
It wasn't designed for that, but you could try to get that to work. If you choose '--tool' to be the malware which requires the components, you could specify the required components (DLLs) with '--tool-dir'. 

On the guest/sandbox the '--tool' and the contents of '--tool-dir' will be placed in the "%temp%/tool" directory and the malware is placed in the "%temp%" directory.

One thing to note is that specifying '--tool' will bypass the cuckoomon dll.

jbremer:
Wouldn't this more appropriate in the form of an `auxiliary` module? As in, create an auxiliary module that starts the tool with certain arguments. You're sort of implementing the same logic as one, but from scratch, due to which a lot of extra code is introduced which would otherwise not be necessary.

botherder:
I agree with @jbremer

depaolism:
Thanks for the feedback! I've begun working on transitioning this into an auxiliary module.

gtback:
Maybe I'm misunderstanding how auxiliary modules are supposed to be used. The current ones (human, screenshot, disguise) all seem to run in parallel to a sandbox execution with cuckoomon.dll. The intent behind the "tools" which would run via the code in this branch is that they would run in place of hooking the process (sort of like the `--free` flag, which was used as the basis for this idea). The idea is to reuse the sandbox management, task management, agent communication, and web/reporting infrastructure of Cuckoo to run things like dynamic unpackers or tracing tools that supplement Cuckoo's "traditional" API hooking via the DLL. In many cases, these would conflict with the API hooking such that you couldn't run them both at the same time on the same sample. But by submitting the sample twice to Cuckoo (once "normally" using cuckoomon.dll) and again with an unpacker or tracing tool, additional information could be learned about the malware. I couldn't find any way to enable or disable auxiliary modules on a per-task basis, only through the config file (which doesn't make sense for submitting the same sample twice). This is why it was implemented using additional arguments to submit.py. 

Please let me know if I'm misunderstanding auxiliary modules, but I don't see how they would be better suited for this than the current approach.

gtback:
@botherder / @jbremer Can you explain how an auxiliary module helps? Many of these tools need to run in the same process as the malware, rather than simply interacting directly with the OS (for example, moving the mouse or taking a screenshot) or a different process.

jbremer:
So auxiliary modules can be used to run tools separately on the system - tools that are not interacting directly with the sample being analyzed. As an example of one, indeed the screenshot module.
In your initial description you mentioned the following: `An example of an analysis tool would be a dynamic unpacker or an execution tracer.`
Now dynamic unpackers and execution tracers will probably work directly on the malware, much like Microsoft Word works directly with .docx documents. Hence I think it'd make more sense to, for this goal, create a new Package (such as https://github.com/cuckoobox/cuckoo/blob/master/analyzer/windows/modules/packages/doc.py), and start the sample from there. You can supply, e.g., `pin.exe` and your specific execution tracer library (pintool) in the `bin/` directory of `analyzer/windows/bin`.

What do you think?

gtback:
@jbremer: I agree, it seems like this "tool" feature duplicates a lot of the existing functionality (passing options to tools, returning files, etc.) already provided by `Package`s. By putting any necessary binary files in the analyzer/windows/bin and passing "free=True" to execute, we should be able to do what we need by writing individual Packages for each tool rather than a Tool package (as we did in #218) or the approach in this PR to add things to the Cuckoo core. 

I have to imagine that either the abstract Package has changed a lot since we started working on this, or I just didn't realize that it was pretty much what we needed all along.

@depaolism : Unless you have any other thoughts, I think we can go ahead and close this.

jbremer:
@gtback @depaolism Ok, glad to see my thoughts came through. Please, if you're working on an execution tracer plugin (as package with the binaries in `analyzer/windows/bin` etc) then I'd be happy to merge that.
In fact, for specific malware this could be very interesting. Do keep in mind that one should be able to specify different pintools (through the `dll` parameter probably).
Actually, writing the Cuckoo part should be pretty straightforward, but the pintool might be a bit more work, and only useful for specific malware (unpackers etc I think).

Issue #511, opened 2015-04-03
Deprecation of web.py and/or HTML reports


botherder:
Let's discuss here what is to do with HTML reports and the `web.py` utility.
As we currently stand, both of them are largely abandoned as all the development is being done on the Django web interface.

The Django web interface provides a much better view of the single reports as well as extended search and comparison capabilities. From our point of view, it is definitely the preferred choice.

Before taking any decision, we should get some feedback from our users on what is the best course of action. HTML reports in particular might still have a significant value, and we maybe we should revive them instead of abandoning them.

jbremer:
Generally I think people would rather see a PDF report than a HTML report. So maybe start deprecating HTML reports and create a PDF reporting module which people can use as replacement. Of course this means actually making a PDF which is kind of a pain (i.e., either use some unknown library or use something bloated like LaTeX).

Maybe an idea to ask people on the mailing list about this issue, @botherder?

gtback:
It might be possible to use something like [`weasyprint`](http://weasyprint.org/docs/), keeping the HTML report and supporting conversion to PDF. I've never used it personally, though.

g-clef:
Right now I'm using both the html and json reports. I have a custom front-end that submits files to both Cuckoo and another malware tester simultaneously & collects the results from both. The Cuckoo json results go to ElasticSearch, and the html results are shown to the user. I could deal with deprecating the html report, since I can always scrape the django front-end if I have to.

One side-note: I pretty strongly disagree with the idea that people would prefer PDFs. I hate them because getting data out of a PDF to present to the user is annoying & embedding a PDF to a user is a bit of a cop-out.

jbremer:
@g-clef You're not supposed to get data out of a PDF nor a HTML document, for that purpose there's the JSON report which contains all the information (if not more) that you can also find in the HTML (and possibly in the future PDF) report.

g-clef:
@jbremer for my purpose, embedding a PDF is a very bad user-interface option. Embedding HTML is fine even if I only rip part of the HTML out & display it to the user. But I at least have that option with HTML, and can preserve some of  the original layout & appearance with HTML. With PDF, I have no choice but to present the user with a foreign thing, breaking the user's interaction with my site (can't preserve visual appearance, can't ensure links back to my system, etc).

(But this is a derail of the original conversation. I'll stop here, and just register my vote that I do use the HTML web.py interface.)

botherder:
[...]

PR #566, opened 2015-06-05
Fix #564 and #511 change utils/web.py to a shortcut of runserver of web in Django

### LINKED ISSUE #564: Proposal: remove utils/web.py web and that it run Django web with development server
That utils/web.py not run other web interface and that run Django development server, for only develop a web user interface. Merge two webs in one.

jhg:
This fix #564 and when HTML reports will be deleted jinja2 can be remove from requirements.txt


jhg:
Now this pull also fix #511

jhg:
This need #578

jbremer:
Replacing the `web.py` by calling into the Django web interface doesn't sound like a very good idea to me. When people run `web.py` they expect to see the `web.py` interface and not our new Django web interface. To me it'd be better just to remove the `web.py` altogether and at the very most leave a script that just prints an error and exits.
Thoughts @jekil @botherder?

jhg:
When decide that do with web.py I can delete it or show a error when run web.py script.

jhg:
@jbremer delete it or show a error when run?

botherder:
We kind of just nuked utils/web.py. I don't think it is necessary to had this kind of trampoline, as long as the changes are appropriately reflected in the documentation.

jbremer:
We're revamping the standalone HTML/PDF report in #863 - part of that being getting rid of these files. So closing this issue for that reason.

PR #568, opened 2015-06-07
Submit sample via email

edwincheese:
- Utility functions to extract attachments from MIME and RFC 2822-based
  message.
- New API endpoint in api.py. The API accepts raw email text in either
  request body or form data (compatible with mailgun)


jbremer:
I like the integration with emails, but what happens when you have an email with multiple attachments?

edwincheese:
@jbremer Currently, multiple samples in an email will be submitted for analysis individually. Also, if the email contains attachments in RFC 2822-like format (e.g. a copy of email source code), it will parse the attachment and submit any attachment inside the attachment. It allows people to submit samples that they received via email using "Forward as Attachment" function in their email client. The raw email source (headers, email body, etc) is not used at this moment, but it could be saved for further analysis as well.

jekil:
I like this PR.

jbremer:
Little bit late to the party, but this functionality has now been included in https://github.com/jbremer/sflock, and will be incorporated in the Cuckoo API and Cuckoo Web Interface in a commit nearby.
Therefore going to close this issue now. Thanks again for the contribution, and sorry for the much belated reaction ;-)

PR #573, opened 2015-06-10
Issue deleting tasks with identical tags

Shane-Carr:
I have found that if several tasks are submitted with a given tag, deleting a single task via the API deletes all of the tasks.

Appears to be an issue with the unique constraint on the tags.names column and the bi-directional relationships between the tasks, tasks_tags and tags tables as defined in database.py.  

Adding two samples with tag 'a':

curl -F file=@/tmp/sample.exe --form_string "tags=a" http://server/tasks/create/file
curl -F file=@/tmp/sample2.exe --form_string "tags=a" http://server/tasks/create/file

sees the following in the DB tables:

**tasks**
id=1, target=sample.exe
id=2, target=sample.exe

**tasks_tags**
task_id=1, tag_id=1
task_id=2, tag_id=1

**tags**
id=1, name=a

Then issuing 'curl http://server/tasks/delete/1', both tasks 1 and 2 are deleted and I am left with:

**tasks**

**tasks_tags**

**tags**

Both tasks have been deleted despite only requesting task 1 to be deleted.

The proposed change addresses the issue.

### DIFF
--- a/lib/cuckoo/core/database.py
+++ b/lib/cuckoo/core/database.py
@@ -114,7 +114,7 @@ class Tag(Base):
     __tablename__ = "tags"
 
     id = Column(Integer(), primary_key=True)
-    name = Column(String(255), nullable=False, unique=True)
+    name = Column(String(255), nullable=False, unique=False)
 
     def __repr__(self):
         return "<Tag('{0}','{1}')>".format(self.id, self.name)
@@ -411,19 +411,6 @@ class Database(object):
                                         "import %s (install with `pip "
                                         "install %s`)" % (lib, lib))
 
-    def _get_or_create(self, session, model, **kwargs):
-        """Get an ORM instance or create it if not exist.
-        @param session: SQLAlchemy session object
-        @param model: model to query
-        @return: row instance
-        """
-        instance = session.query(model).filter_by(**kwargs).first()
-        if instance:
-            return instance
-        else:
-            instance = model(**kwargs)
-            return instance
-
     @classlock
     def drop(self):
         """Drop all tables."""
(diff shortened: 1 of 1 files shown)


jbremer:
Very interesting issue. I'll have to confirm the bug later, thanks.

gtback:
This sounds like a cascading delete issue. I'm not sure if the right approach is to create multiple tags with the same name (which might hamper searching, and is the point of `tasks_tags` "joing" table) vs. resolving the issue that deleting a task also deletes tags associated with that task, even if there are other tasks using the same tag.

Shane-Carr:
I'd certainly be interested to know if there is an alternate solution that maintains the uniqueness in the tags table.  

Perhaps @jekil has some thoughts given he authored that part of the DB schema?

Shane-Carr:
I have done some more testing of this and found a solution that addresses the initial bug but still maintains the uniqueness on the tags table.  Any orphaned tags are automatically cleaned up once they are no longer referenced by any machines or tasks.

This is based off the example provided here:

http://stackoverflow.com/questions/9234082/setting-delete-orphan-on-sqlalchemy-relationship-causes-assertionerror-this-att/9264556#9264556

jbremer:
I've recently run into the same issue. Given this issue is around `tags` though I'd be in favour of getting rid of the whole `cascade="all, delete"` at all - in practice there are only, at the very most a few dozen tags anyway.
Getting rid of it also fixes this issue. Thoughts?

Shane-Carr:
I believe that will work OK as well.  It just means that every time a user submits a task and typos the tag or uses a non-existent tag, it will end up left in the tags table.  That was why I proposed the event listener to clean up any orphans along the way.  Really shouldn't be a big deal though I guess.

jbremer:
Sorry for the delay. I've just committed the patch that I proposed upstream. In this case I prefer simplicity over relatively complex logic with the event listeners etc.
Anyway, closing this issue as it is resolved for me. Please re-open or create a new issue if you're still experiencing issues.. :)

PR #576, opened 2015-06-10
Consolidate ESX connections to prevent connection spam

carthurb:
In production/test environments we have run into issues with connections to the ESX server using libvirt failing due to 500/503 errors.  Each cuckoo initialization rapidly opens and closes (number_of_machines * 2) + 1 connections, and each time a submission is sent to all machines rapidly opens and closes number_of_machines connections.

We aren't entirely sure if the issue stems from an issue closing sessions on ESX, with libvirt, or with cuckoo, but this change consolidates to a single connection to the server which reconnects on timeout and results in 1 open connection per cuckoo instance regardless of the number of analysis/machines being run.  This has resolved the 500/503 connection issues.


botherder:
Thanks for the notification. Please, @jekil can you look into it?

KillerInstinct:
This works in my setup.

jbremer:
Looks

PR #590, opened 2015-06-26
Fix shallow copy bug in MongoDB reporting module

fedelemantuano:
MongoDB reporting module makes a shallow copy of `results` dict instead of a deep copy.
This behaviour can affect other reporting modules.

### DIFF
--- a/modules/reporting/mongodb.py
+++ b/modules/reporting/mongodb.py
@@ -3,6 +3,7 @@
 # See the file 'docs/LICENSE' for copying permission.
 
 import os
+import copy
 
 from lib.cuckoo.common.abstracts import Report
 from lib.cuckoo.common.exceptions import CuckooDependencyError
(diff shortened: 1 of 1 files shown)


botherder:
Bizarre, we had this problem before and I thought it was fixed. Thanks for reporting.

jbremer:
We have reverted this commit due to #591. Can you share an example reporting module that causes the bug for which you provided a workaround, @fedelemantuano?

fedelemantuano:
Hi @jbremer I created a report module for Elasticseach and I had some problems with nested fields.

jbremer:
Hi @fedelemantuano, could you please provide us with analysis log (i.e., the full analysis directory of your task) with expected input and output for the various modules that you mention.
Also, assuming you'll eventually PR the Elasticsearch module, could you please show us the current implementation of your module? Personally I have interest in this module - so if you're not interested in sharing it back with the community I'll probably develop it myself ;)

fedelemantuano:
Hi @jbremer, I'm reverting mongodb.py module and I will provide you some analysis log.
Yes, I will share with you this module with a pull request so we can work together.

PR #592, opened 2015-06-30
Fixed path for Office 15 (Office 2013) Home/Business.

nsmfoo:
The "auto installer" puts Office 2013 in this path. As I'm unsure of the origin of the current path, I kept it for you to decide if it stays or not, as the path might differ ..


jbremer:
Please delete the whitespace from `Program Files` - this is a magic identifier. Now it'll probably not spew out the correct paths, if anything useful at all.

PR #598, opened 2015-07-07
Update esx.conf

dewiestr:
this is identical for all the virtual-conf files. If the interface is set it will override the default interface specified in AUXILIARY.CONF and not the cuckoo.conf ( recently changed )


jbremer:
Oh, already took care of this in fc57b063ef37de11d3eb9d1abc41a37533f629ba as part of #597.

PR #604, opened 2015-07-22
Alternative machinery module for vSphere/ESXi with memory-dump support

jgajek:
This is a vSphere/ESX machinery module for the Cuckoo sandbox and is intended as an alternative to Cuckoo's ESX machinery module.

It adds full memory dump functionality during analysis, which is currently not supported by Cuckoo's LibVirt-based ESX module.

The module depends on the vSphere API Python Bindings provided by VMware (https://github.com/vmware/pyvmomi) which can be installed via pip:

sudo pip install --upgrade pyvmomi

### DIFF
--- a/conf/vsphere.conf
+++ b/conf/vsphere.conf
@@ -1,0 +1,58 @@
+[vsphere]
+
+# ESXi host connection parameters
+host = 10.0.0.1
+port = 443
+user = username_goes_here
+pwd = password_goes_here
+
+machines = analysis1
+
+[analysis1]
+# Specify the label name of the current machine as specified on your
+# vSphere host.
+label = cuckoo1
+
+# Specify the operating system platform used by current machine
+# [windows/darwin/linux].
+platform = windows
+
+# Please specify the name of the snapshot. This snapshot should be taken
+# while the machine is running and the agent started.
+snapshot = cuckoo_ready_running
+
+# Specify the IP address of the current virtual machine. Make sure that the
+# IP address is valid and that the host machine is able to reach it. If not,
+# the analysis will fail.
+ip = 192.168.1.1
+
+# (Optional) Specify the name of the network interface that should be used
+# when dumping network traffic from this machine with tcpdump. If specified,
+# overrides the default interface specified in auxiliary.conf
+# Example (eth0 is the interface name):
+# interface = eth0
+
+# (Optional) Specify the IP of the Result Server, as your virtual machine sees it.
+# The Result Server will always bind to the address and port specified in cuckoo.conf,
+# however you could set up your virtual network to use NAT/PAT, so you can specify here
+# the IP address for the Result Server as your machine sees it. If you don't specify an
+# address here, the machine will use the default value from cuckoo.conf.
+# NOTE: if you set this option you have to set result server IP to 0.0.0.0 in cuckoo.conf.
+# Example:
+# resultserver_ip = 192.168.122.101
+
+# (Optional) Specify the port for the Result Server, as your virtual machine sees it.
+# The Result Server will always bind to the address and port specified in cuckoo.conf,
+# however you could set up your virtual network to use NAT/PAT, so you can specify here
+# the port for the Result Server as your machine sees it. If you don't specify a port
+# here, the machine will use the default value from cuckoo.conf.
+# Example:
+# resultserver_port = 2042
+
+# (Optional) Set your own tags. These are comma separated and help to identify
+# specific VMs. You can run samples on VMs with tag you require.
+# Note that the 64_bit tag is currently special.  For submitted 64-bit PE files,
+# the 64_bit tag will automatically be added, forcing them to be run on a 64-bit
+# VM.  For this reason, make sure all 64-bit VMs have the 64_bit tag.
+#tags = windows_xp_sp3,32_bit,acrobat_reader_8
+
--- a/modules/machinery/vsphere.py
+++ b/modules/machinery/vsphere.py
@@ -1,4 +1,6 @@
 # Copyright (C) 2015 eSentire, Inc (jacob.gajek@esentire.com).
+# This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
+# See the file 'docs/LICENSE' for copying permission.
 
 import requests
 import logging
@@ -20,6 +22,7 @@ except ImportError:
     HAVE_PYVMOMI = False
 
 log = logging.getLogger(__name__)
+logging.getLogger("requests").setLevel(logging.WARNING)
 
 
 class vSphere(Machinery):
@@ -33,7 +36,8 @@ class vSphere(Machinery):
 
     def __init__(self):
         if not HAVE_PYVMOMI:
-            raise CuckooDependencyError("Couldn't import pyVmomi")
+            raise CuckooDependencyError("Couldn't import pyVmomi. Please install "
+                                        "using 'pip install --upgrade pyvmomi'")
 
         super(vSphere, self).__init__()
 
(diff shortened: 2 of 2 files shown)


jbremer:
Looks good to me - thanks. Will need somebody to confirm all works fine though (just to make sure).

jgajek:
No problem Juriaan.  I've been using it successfully for 2-3 months now, but mostly with Brad Spengler's modified Cuckoo build, so additional validation can't hurt.

KillerInstinct:
I'll test this tomorrow. Does this work well with volatility profiles for Vmware Snapshot memory dumps?

jgajek:
It should.  I tested it with Cuckoo's memory analysis module backed by Volatility 2.4 and it worked fine.

KillerInstinct:
Neato, I believe I am running 2.3.x -- will check then. :)

KillerInstinct:
I tested this with a few analyses now and it works as expected. I do have a few suggestions for changes though:

1) Squelch requests logs by adding the below code:
```
logging.getLogger("requests").setLevel(logging.WARNING)
```
Otherwise we see each HTTPS API request to the ESXi/vSphere server.

2) Replace the error handling when the config file doesn't have values. EG if I didn't specify a user in the config file, it will default to try to log in with root. I think this should be replaced with CuckooCriticalError(). I would do this for the port too. (You already do this for Host/Password config values anyways.)

3) Minor grammar fixes:
* In line 53 of the machinery file you state you are raise'ing CuckooMachineError when in fact you raise CuckooCriticalError
* If we don't have pyVmomi, the error message on line 36 doesn't have the "use pip install pyvmomi to install" message which is in other places of Cuckoo.

Going to let this burn in some, but functionally it works the same as esx machinery module, and this has the benefit of not relying on libvirt to update their code. EG I'd expect this to work with vSphere 6.x where libvirt is does not support this version.

Overall, it seems it loads slightly slower than libvirt, but its almost unnoticeable. I'm okay with that considering it produces less logs and has the ability to grab snapshots. However with Volatility 2.3.1 I am unable to get it to parse:

(Snipped this to relevant logs)
```
AddrSpaceError: No suitable address space mapping found
VMWareSnapshotFile: No base Address Space
VMWareSnapshotFile: Invalid VMware signature: -
```

Going to give volatility 2.4 a shot though. Anyways, will report back in a week or so once I have some bandwidth to see how it handles 5-8 VM's running at the same time. If all goes good I'd honestly suggest replacing the current esx machinery with this one.

PS: I tested on ESXi 5.5.

KillerInstinct:
Oh and also, you have configuration options for memory profiles in the vcenter config, which is not yet supported by upstream. (I presume they are dev'ing that now based off the open PR's)

KillerInstinct:
I'm not sure what the RNG is used for in the _dump_memory function. Seems like its being used to set the name of the snapshot, but ESXi appears to just name the snapshot +1 of the previous.

See below (I printed out the request arguments):
```
2015-07-23 09:28:25,562 [modules.machinery.vsphere] INFO: Downloading memory dump [Array01] Cuckoo01/Cuckoo01-Snapshot15.vmsn to /opt/cuckoo/storage/analyses/778/memory.dmp
{'dsName': 'Array01'}
{'Cookie': 'vmware_soap_session="52608b7e-3378-d1c2-72b1-fe70b0963367"; Path=/; HttpOnly; Secure;'}
https://[IP Address]:443/folder/Cuckoo01/Cuckoo01-Snapshot15.vmsn
2015-07-23 09:33:14,999 [modules.machinery.vsphere] INFO: Removing snapshot cuckoo_memdump_408 for machine Cuckoo01
```

jgajek:
Thanks for the detailed feedback, Jeremy.  I will try to get your suggestions implemented and checked in today.  With regards to the random number generator -- it is just used to generate a snapshot name (cuckoo_memdump_XXX) to reduce the chance of naming conflicts on the vSphere server.

Regarding loading times, yes I would expect this one to load a bit slower.  It does some additional initialization checks to make sure that the snapshots for all defined VM guests were taken in powered on state.  The original ESX machinery module will just hang and time out if a powered off snapshot is used, which can be confusing the first few times you are bitten by this issue.

KillerInstinct:
@jgajek The snapshot that is saved in my datastore does not get saved as the RNG name, it gets saved as Cuckoo01-Snapshot14, to which the next one would be saved as Cuckoo01-Snapshot15. Does yours actually save that name to the datastore? If so, what version of ESXi are you using?

jgajek:
The files are named sequentially as you describe, but there is also a label attached to each snapshot.  You should be able to see the labels in Snapshot Manager.  The memory dump snapshots are deleted by the machinery module after downloading, so the chance of naming conflicts is small, but it is just a "just in case" thing.  Since I retrieve snapshots by label in the code, I wanted to avoid ambiguities as much as possible.

KillerInstinct:
Ahh that make sense, in that case, I would suggest increasing the RNG, as setups with 50+VM's may have a higher chance of collision. I'd suggest 10000000-99999999 or something similar.

jgajek:
Done.

jbremer:
Thanks @jgajek and @KillerInstinct for the feedback!

vpistillo:
When I run cuckoo I receive always the some error also if I run the command suggested (./utils/community.py -wafb monitor).

 2015-09-01 14:55:49,992 [root] WARNING: The binary analyzer/windows/bin/monitor-x86.dll is more than a week old! 2015-09-01 14:55:50,031 [root] WARNING: The binary analyzer/windows/bin/monitor-x64.dll is more than a week old!

2015-09-01 14:55:50,039 [root] WARNING: The binary analyzer/windows/bin/inject-x86.exe is more than a week old! 2015-09-01 14:55:50,048 [root] WARNING: The binary analyzer/windows/bin/inject-x64.exe is more than a week old!

2015-09-01 14:55:50,057 [root] WARNING: The binary analyzer/windows/bin/is32bit.exe is more than a week old!

2015-09-01 14:55:50,057 [root] CRITICAL: It is recommended that you update the binaries used for Windows analysis (if you have not done so already, it is possible that there was no update - in that case this error will persist). To do so, please run the following command: ./utils/community.py -wafb monitor.

vpistillo:
When I run cuckoo I receive also the following INFO:

2015-09-02 11:22:04,404 [lib.cuckoo.core.scheduler] INFO: Using "vsphere" machine manager with max_analysis_count=0, max_machines_count=0, and max_vmstartup_count=10
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.

  InsecurePlatformWarning
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/connectionpool.py:768: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)

vpistillo:
During the malware analysis I receive the following message:

2015-09-01 12:29:06,646 [modules.processing.dumptls] INFO: TLS Master secrets file not found for analysis

jbremer:
@vpistillo Most of those are legitimate errors, but I'll try to make them more user friendly.

jbremer:
@jgajek I may have been a bit prosperous with merging, we have a small follow-up question from the Cuckoo Foundation point of view, namely if you could sign a CLA for us. Feel free to email us at cuckoo@cuckoofoundation.org for more information and/or please check http://cuckoofoundation.org/contribute.html - thanks!

jgajek:
Jurriaan, I will check with our senior management about signing the
agreement.

PR #534, opened 2015-04-28
Added datetime field in mongodb report to perform a query date in mongodb.

fedelemantuano:
Added datetime field in mongodb report: report["info"]["started_datetime"].
Now it's possible to perform a query date in mongodb with pymongo.

Query example:

results = results_db.analysis.find({"$and": [{"info.started_datetime":{"$gte": from_date}}, {"signatures.description": SIGNATURE}]})

where from_date is a datetime type.

PR #606, opened 2015-07-24
Elasticsearch report module

fedelemantuano:
Elasticsearch report module as discussed in issue #590 with @jbremer.

### DIFF
--- a/conf/reporting.conf
+++ b/conf/reporting.conf
@@ -32,3 +32,10 @@ port = 27017
 db = cuckoo
 store_memdump = yes
 paginate = 100
+
+[elasticsearchdb]
+enabled = no
+hosts = 127.0.0.1:9200
+index = cuckoobox
+doc_type = analysis
+timeout = 30
--- a/modules/reporting/elasticsearchdb.py
+++ b/modules/reporting/elasticsearchdb.py
@@ -1,0 +1,192 @@
+# Cuckoobox to ELK
+
+# -*- coding: utf-8 -*-
+
+import datetime
+import copy
+import logging
+
+from lib.cuckoo.common.abstracts import Report
+from lib.cuckoo.common.exceptions import CuckooDependencyError
+from lib.cuckoo.common.exceptions import CuckooReportError
+
+try:
+    from elasticsearch import Elasticsearch
+    from elasticsearch import ConnectionError
+    from elasticsearch import ConnectionTimeout
+    HAVE_ELK = True
+except ImportError:
+    HAVE_ELK = False
+
+
+log = logging.getLogger(__name__)
+
+
+class ElasticsearchDB(Report):
+    """Stores report in Elasticsearch."""
+
+    def get_options(self, options_string):
+        """Get analysis options.
+        @return: options dict.
+        """
+        # The analysis package can be provided with some options in the
+        # following format:
+        #   option1=value1,option2=value2,option3=value3
+        #
+        # Here we parse such options and provide a dictionary that will be made
+        # accessible to the analysis package.
+        options = {}
+        if options_string:
+            try:
+                # Split the options by comma.
+                fields = options_string.strip().split(",")
+            except ValueError as e:
+                log.warning("Failed parsing the options: %s", e)
+            else:
+                for field in fields:
+                    # Split the name and the value of the option.
+                    try:
+                        key, value = field.strip().split("=")
+                    except ValueError as e:
+                        log.warning("Failed parsing option (%s): %s", field, e)
+                    else:
+                        # If the parsing went good, we add the option to the
+                        # dictionary.
+                        options[key.strip()] = value.strip()
+
+        return options
+
+    def walk_dict(self, d, k=None, dk=None):
+        if type(d) == type({}):
+            for k in d:
+                self.walk_dict(k=k, dk=d, d=d[k])
+        else:
+            if k:
+                if isinstance(dk[k], str):
+                    try:
+                        dk[k] = unicode(dk[k], errors='ignore').encode(
+                            encoding='UTF-8')
+                    except:
+                        log.exception("Except in instance str")
+                        dk[k] = ''
+
+                elif isinstance(dk[k], unicode):
+                    try:
+                        dk[k] = dk[k].encode(encoding='UTF-8')
+                    except:
+                        log.exception("Except in instance unicode")
+                        dk[k] = ''
+                else:
+                    pass
+
+    def connect(self):
+        """Connects to Mongo database, loads options and set connectors.
+        @raise CuckooReportError: if unable to connect.
+        """
+        hosts = self.options.get("hosts", "127.0.0.1:9200").split(",")
+        log.info("Elasticsearch hosts: {}".format(hosts))
+        index = self.options.get("index", "cuckoobox")
+        log.info("Elasticsearch index: {}".format(index))
+        doc_type = self.options.get("doc_type", "cuckoo_type")
+        log.info("Elasticsearch doc_type: {}".format(doc_type))
+        elk_timeout = self.options.get("timeout", "10")
+        log.info("Elasticsearch create timeout: {} s".format(elk_timeout))
+
+        try:
+            self.es = Elasticsearch(hosts)
+            self.index = index
+            self.doc_type = doc_type
+            self.elk_timeout = elk_timeout
+        except TypeError:
+            raise CuckooReportError("Elasticsearch connection hosts must be host:port or host")
+        except ConnectionError:
+            raise CuckooReportError("Cannot connect to Elasticsearch (ConnectionError)")
+        except ConnectionTimeout:
+            raise CuckooReportError("Cannot connect to Elasticsearch (ConnectionTimeout)")
+
+    def run(self, results):
+        """Writes report.
+        @param results: analysis results dictionary.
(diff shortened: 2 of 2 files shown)


jbremer:
Thanks for the initial PR @fedelemantuano. I see various improvements to be done here, though.
The `get_options` function is not used and thus can be removed completely.
The `walk_dict` is *fixing* some issues which should be fixed elsewhere. If there are problems with unicode strings then these most likely also pose a problem in the `jsondump` module, etc. It would be much preferred, by me anyway, to get rid of the `walk_dict` logic and fix the problems at the root cause.
Then, as @KillerInstinct pointed out to me a small while ago, doing `copy.deepcopy` on the results is pretty expensive. Imagine having 200mb of data in there ;)
Finally the modifications on the results itself. Like the `walk_dict` stuff I think all of these changes should be done elsewhere, i.e., probably in the `processing` stage, if at all. What's the reason behind the `datetime.strptime` thing? How does ElasticSearch translate this when importing it?

I hope not to sound too harsh or anything - just trying to get the most out of this PR! :)

KillerInstinct:
@jbremer the datetime.strptime is used to brad a date format so that the data is stored in an index based on the started_on date. This allows ES to create index based on date which makes searching much faster, and scaleable with large datasets.

fedelemantuano:
Hi @jbremer. Yes we can remove get_options because I don't use this function here.
I know that `copy.deepcopy` can be a problem, but in ELK module I have to change results without impact others report modules.

I do some modifications to results:
1) **virustotal**: I changed the result for solve this issue https://www.elastic.co/guide/en/elasticsearch/guide/current/complex-core-fields.html#object-arrays.
So we can search for key. For instance: `virustotal.scans.antivirus`.

2) `datetime.strptime`: in ELK is a datetime, so it's possibile search for date and not string (for instance range query). Sorting in ELK is very expensive, but if you use datetime field you can use _doc_value_ that don't use java heap (https://www.elastic.co/guide/en/elasticsearch/guide/current/doc-values.html, https://www.elastic.co/guide/en/elasticsearch/guide/current/preload-fielddata.html).

3) I added tags in results, for use filters in ELK (more performance). For instance: all analysis with network analysis, etc.

Don't worry not to sound too harsh ;)

jbremer:
@KillerInstinct could you please provide me with an answer to the following question: would it be beneficial to use `datetime` objects in the `mongodb` backend as well?
If so I'd suggest that we switch to the usage of `datetime` objects instead of strings everywhere. In reporthtml/jsondump it's trivial to switch and that way no reporting module will have to change the results dictionary.

Other than that I like your VirusTotal & Tags additions - like the other stuff I'm oke moving that higher up (i.e., in a processing module) so that it in theory it can also be used by other reporting modules, but more specifically `mongodb` gets the same information for free.

fedelemantuano:
Hi @jbremer I have done PR #534 to perform a query date in mongodb with pymongo.

jbremer:
Excellent, yes. We should probably still go down the way of doing this through a processing module, though :) Probably adjust the `analysisinfo` one.

KillerInstinct:
@jbremer : Sure it would be beneficial for searching as already observed. I'm no expert at Mongo, but I assume if we ever got around to tuning/indexing/sharding we could do that based on datetime.

jbremer:
That's what I thought. Please allow me a bit of time (due to travelling etc up to a week or two) to properly reflect these changes. After that we update the elasticsearch module to reflect those changes and we're good to go. Deal? If somebody has the time to do this for me that's also appreciated of course!

jbremer:
Sorry for the delay - a lot is happening. I'll be moving forward on this issue soon.

fedelemantuano:
Hi @jbremer. I can help you, but I need to see the new process modules or I need know what do you want to change.

jbremer:
@fedelemantuano We want to have less date strings and more `datetime` objects. But this also involves modifying some `processing` modules and the other `reporting` modules. So needs a bit of testing etc.
That, including moving the `VirusTotal` magic to the `VirusTotal` module, will allow the `ElasticSearch` module to do pretty much nothing but the `self.es.index()` part :-)

jbremer:
Oh, and now I take another look at it, we also have to resolve this issue: https://github.com/cuckoobox/cuckoo/pull/606/files#diff-3cd940e5814382686513f88a7a4c5894R126 ;-)
One thing at a time!

KillerInstinct:
For what its worth, the bson calls should be serializable if you index the processed results. This would have to be redone to your behavior processing, but feel free to use as a guideline:

https://github.com/KillerInstinct/cuckoo-modified/blob/patch-61/modules/reporting/elasticsearchdb.py#L63

jbremer:
That looks exactly the same as what's happening in the MongoDB reporting module, but yeah, thanks ;)

botherder:
Updates on this PR?

fedelemantuano:
Sorry I'm waiting @jbremer

jbremer:
I've reimplemented parts of this PR in a relatively simplistic way, see also https://github.com/cuckoosandbox/cuckoo/blob/master/modules/reporting/elasticsearch.py.
The timestamp stuff has not been ported over yet - as discussed earlier I'd preferably move those changes (from string representations to `datetime.datetime` objects) to the Cuckoo core, so they apply to all processing, signature, and reporting modules, rather than having specific behavior here and there.

Other than that I have yet to read up on ElasticSearch's `tags` functionality and, although I can imagine their purpose, it would help if people could share examples where tags come in handy in practice.

fedelemantuano:
Hi @jbremer. The purpose of tags is use filter query in Elasticsearch. Example: Give me all analysis with tag Zeus. Maybe it's better don't fill tags array in this specific module, but to give to users the possibility to tag the specific task in signature module, when the analysis is terminated.

I see the new module, in https://github.com/cuckoosandbox/cuckoo/blob/master/modules/reporting/elasticsearch.py#L51
you create a new index for every task. Are you sure do you want it?
For me it's better, to manage ELK index with timestamp (purge, close index, etc.) using [curator](https://github.com/elastic/curator).

jekil:
I am also unsure about having an index for each task, imho indexes are meant to be used in a different way. See http://elasticsearch-users.115913.n3.nabble.com/How-many-indexes-is-too-many-indexes-td4064599.html

jbremer:
I was thinking the same as @jekil, but not sure what the *correct* approach should be - it's not definitive yet either. From what I understood @fedelemantuano's indexes were based on the day of the year, so if you analyze 1000 analyses on a particular day, all of those go in one index, right?
Surely there must be a best way to do this properly - anyone?

fedelemantuano:
In Elasticsearch index is only a logical namespace that point to one or more phisical shards. A shard is a low-level worker unit that holds all the data. So in big installation is possible to settings more shards for a single index, but 1000 analyses is very small number.
Elasticsearch suggest to use timestamp index to manage the cluster and to use Kibana GUI.
I thinking to purge old analysis, close indexes, backup, etc.

jbremer:
Yep. But then one index a day doesn't make sense per se either. Keep in mind that generally speaking there are two use cases for Cuckoo; CERT mode (<100 analyses a day) and mass mode (50k+ analyses a day).
Perhaps we should have a sort of rate limiting approach that applies to both situations. That is, switch to a new index when either we enter a new week (so at the very least every 7 days), or when we hit, say, 100k analyses. But then for the last part some sort of state should be kept.. meh.
And this limit also depends on whether a user index just the behavioral *summary* or whether a user also indexes all the behavioral calls; the difference between those two is on a scale of 10000x (think 100mb of behavioral calls vs a 10kb behavioral summary).

KillerInstinct:
@jbremer In talks with the people that wrote most of the code for the fork, they also suggested it would be best to index on both Task ID and date. Indexes really depend on what your needs are. For example if you want to speed up the searching, you'd index the behavior summary keys, VT results, hashes, etc.

I know Metcalf indexed mongo for suricata stuff so that he could better search on Suricata alert definitions because the default mongo search was slow. So -- it's really all based on need and what you guys think is best for performance.

fedelemantuano:
I suggest time based indexes (for day, for months, etc.) because is more efficient the search and the maintenance. 

ELK doesn't have problem to manage 10, 100, 1000, etc. documents. If you are a CERT your index will have a shard or if you are Malwr will have 50 shards.
You can think to do this setting a parameter in report conf. Example `prefix = cuckoo-, timebase = "%Y.%m.%d"`

It's possible to have only one logic index with the use of `alias`.

Some ELK link:
https://www.elastic.co/guide/en/elasticsearch/guide/current/time-based.html
https://www.elastic.co/guide/en/elasticsearch/guide/current/multiple-indices.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-aliases.html

jbremer:
Sounds good to me. Let's predefine a couple of values though (i.e., `day` maps to `%Y.%m.%d` and then `week` and `month`) so users won't shoot themselves in the foot. And then default to the `month` option?
Anything else before we finish off on ES integration?

botherder:
"or if you are Malwr will have 50 shards".
Ha, right, @jekil?

jekil:
@botherder I see shards, shards everywhere.

botherder:
Can we close this pull request?

fedelemantuano:
For me yes because @jbremer has reimplemented module: https://github.com/cuckoosandbox/cuckoo/blob/master/modules/reporting/elasticsearch.py

PR #629, opened 2015-08-26
Fix of TypeError in scheduler

m-messiah:
Bad default None raise TypeError in ",".join(None) (line 231).


jbremer:
The actual error here is the fact that I accidentally use the variable `exports` twice. Will fix properly in a bit - sorry for the inconvenience.

m-messiah:
Confirm as fixed

PR #633, opened 2015-08-28
fix virustotal send file to analysis

m-messiah:
There is option "scan" for VirusTotal in current version of Cuckoo, but it does not work, always return "No antivirus signatures available." if file is new.

These changes fix it, returning url for analysis


jbremer:
Did anyone test this PR? Looks good to me!

jbremer:
Actually, yeah, that sounds about right. Merging it. Thanks for the contribution!

PR #644, opened 2015-09-11
fixed bson importing issue

PickmanSec:
Please enter the commit message for your changes. Lines starting

### DIFF
--- a/analyzer/darwin/lib/core/host.py
+++ b/analyzer/darwin/lib/core/host.py
@@ -6,8 +6,8 @@
 import yaml
 import socket
 import logging
+import bson
 from os import path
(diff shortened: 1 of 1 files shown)


KillerInstinct:
What exactly is the issue? IIRC you should be using Pymongo's version of bson.

jbremer:
Sorry, missed this PR before. @PickmanSec you do realize this does not actually fix anything, it is merely cleaning up the code, right? ;-)

KillerInstinct:
Ahh derp, misread the file path. :)

jbremer:
Still don't think the title represents the PR, but I like it nonetheless. Thanks!

PR #651, opened 2015-09-30
RAR module

m-messiah:
Implement RAR analyzer module for unpacking rar archives as ZIP.
Require installing rarfile into guest windows system, and unrar.exe in $PATH


jbremer:
The `rarfile` package is required in the guest, so adding it to `requirements.txt` doesn't help a lot. Other than that looks fine to me.

botherder:
As usual, I'm against adding additional dependencies inside the guests.

m-messiah:
Do you have any suggestions how to prevent it, but still be able to handle rar packages? Unpacking at server-side?

jbremer:
We need to abstract away the process of analyzing and unpacking zip files in the guest. Preferably a 3rd party project such as Viper (http://viper.li/) or SFlock (https://github.com/jbremer/sflock) would be useful for this stage - determining how many analyses should be created from a zip/rar/7z/etc and which files should be embedded in those analyses.
E.g., a zip with 5 executables in it. Would the best solution be to distribute all 5 samples to 5 different analyses, but in each analysis execute a different executable? Or execute all 5 executables in one analysis? Or.. so many choices here, unfortunately.

jbremer:
We're currently working on this functionality but in a much more broad way allowing better analysis options, so going to close this PR. Thanks though!

PR #652, opened 2015-10-04
Feature class view and login required

jhg:
It now use Django database for auth, but is possible develop a custom login for auth user and password in Cuckoo API without database in Django


jbremer:
To be honest I had expected a bit more from this PR. It seems like instead of having a function for each route, we now have a class and a `get` or `post` function. To me this doesn't seem like an improvement, really.
Could you maybe do a separate PR for just the authentication part? I assume this should also be possible with the current Web interface, right?

jbremer:
For reasons mentioned in my previous comment I don't think there's much to gain from this change, except for an extra indent everywhere. So going to close it, sorry..

jhg:
@jbremer It's a fast PR, for do that you expected I need more time when I have more time. :D I can continue my branch for do it better and re-PR if you look good it. :)

jhg:
@jbremer can you re-open this and assign to me please? I don't know when I can finish this but I can continue this when I will have time

PR #656, opened 2015-10-16
Add workaround for PEP-0476 SSL certificate verification issues

jgajek:
Recent Python versions (2.7.9+) implement PEP-0476, which enables strict SSL certificate verification by default in the standard ssl module. The ssl module is being used internally by the VMware Python SDK, which has not yet been updated to allow unverified SSL connections via it's Connect() and SmartConnect() interfaces. There is no easy way to avoid the resulting connection failures in environments that use a self-signed certificate on the ESXi host.

This pull request adds a 'unverified_ssl' parameter to the vsphere.conf file which globally turns off SSL verification in the ssl module as a workaround.


jbremer:
I feel like this potentially gets rid of all the verified SSL traffic, e.g., also HTTPS traffic emitted by the VirusTotal processing module. Any thoughts on that matter?

jgajek:
Hi Jurriaan,

Any modules that rely on the standard Python ssl module would be affected.  This workaround is intended as a last resort for those who don't have the option to install a proper SSL certificate on their malware lab ESXi host.  Note that I have tried some less drastic options, like adding the default VMware self-signed certificate to the system trusted certificate store, but to no avail.  Turning off verification globally in the process was the only reasonable workaround I could find.  Once the VMware Python SDK folks get around to implementing a proper solution for this (according to the Github discussions on their repo they are working on it), I will update this machinery module so that the 'unverified_ssl' option only affects SSL connections to the ESXi host.

jgajek:
Jurriaan,

the latest version of the pyvmomi SDK (v6.0.0) now has the capability to pass in a SSL context to the connection, so it is no longer necessary to globally disable certificate checking in the entire process.  I have update the vSphere machinery module to take advantage of this.  The 'unverified_ssl' option will now only affect connections to the vSphere host originating from the machinery module and nothing else.

jbremer:
Sorry for the late response, should this PR still be merged?

jgajek:
Yes, this is still relevant.

jbremer:
Thanks again @jgajek, merged it (finally).

PR #662, opened 2015-10-18
Update agent.rst

flo-vienna:
Updated infos how to mount the GuestAdditions.iso with cmd-line online virtualbox for installation of the Guest Additions in order to be able to use transient shares and other features.


jbremer:
We actively pursue people *not* to install the VirtualBox Additions in the VM, as malware is quite good at detecting it (and doing an early exit). So this documentation is probably not so useful :-)
Thanks for the contribution but afraid I'm going to have to close it.

flo-vienna:
Hi

I 100% understand the problem, yet I had to install them to be able to 
update the agent.py file in my very specific setup, but you are right, I 
guess I should uninstall them again after updating the agent.py file :)

Br

jbremer:
I have the feeling when you're uninstalling you're already too late. If your VM has (temporary) access to the internet, you can just download it from github.com :-)

Jurriaan

PR #666, opened 2015-10-23
Added option to report to retrieval to return json from mongodb

trogdorsey:
Instead of storing data in mongodb and on disk, this option will return a json object from the data in mongodb.

### DIFF
--- a/utils/api.py
+++ b/utils/api.py
@@ -18,11 +18,19 @@ except ImportError:
 
 sys.path.append(os.path.join(os.path.abspath(os.path.dirname(__file__)), ".."))
 
+from lib.cuckoo.common.config import Config
 from lib.cuckoo.common.constants import CUCKOO_VERSION, CUCKOO_ROOT
 from lib.cuckoo.common.utils import store_temp_file, delete_folder
 from lib.cuckoo.core.database import Database, TASK_RUNNING, Task
 from lib.cuckoo.core.database import TASK_REPORTED, TASK_COMPLETED
 from lib.cuckoo.core.startup import drop_privileges
+
+HAS_MONGO = True
+try:
+    import pymongo
+    import bson.json_util
+except ImportError:
+    HAS_MONGO = False
 
 # Global Database object.
 db = Database()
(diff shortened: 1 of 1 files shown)


jbremer:
Anyone have thoughts on this matter? Yes/no?

sooshie:
FWIW I like it.

botherder:
I would rather have the interface retrieve the original report.json file instead of regenerating it.

KillerInstinct:
And if you have report.json turned off? Some people prefer to save storage space by not having reports turned on -- the additional option is simple to turn on/off.

jbremer:
Maybe this would make more sense in the context of an API directly embedded in the Cuckoo Web interface (where access to `report.json` may simply be impossible due to MongoDB running on a different host) - something that @KillerInstinct has already worked on in his fork I believe.
Due to the Cuckoo API being a standalone thing from the web interface I feel like this may quickly get outdated when updates or additions will come through.

trogdorsey:
@KillerInstinct is right on the motivation for the modification.  And to request the data straight from the db, you would have to specify that, by default it would grab the report.json.

jekil:
I don't like the idea of having web interface fetching data from storage directory. I think web interface should be able to be decoupled from a "running" cuckoo and be installed in another host, which acts only as web server. So the only common point is mongodb.
I like this contribution.

jbremer:
For reasons I mentioned above I don't think it's a good thing to merge this PR. If you need the JSON report you will/should probably know about that beforehand.
I'd be up for an alternative approach where the Web Interface somehow talks to the Cuckoo API (which may be running on a separate machine - so to allow having the Cuckoo node and the Cuckoo Web Interface separated) and then loads up the original report, assuming it hasn't been removed yet. This requires quite a lot of changes though, but I feel that'd be a more appropriate step forward.

Please feel free to share your thoughts and/or wishes here. Otherwise I'll probably close this PR in the future.

jbremer:
Thanks for your contribution, but as stated earlier I don't think this is going to work out (at least for me - and I'll probably end up maintaining it otherwise anyway). If anybody thinks otherwise, please feel free to re-open the issue. Closing as I currently have no intention of merging this.

PR #678, opened 2015-11-08
Initialize network when running on Virtualbox machinery

lehmz:
This pull request adds functionality to automatically initialize interfaces associated with the ResultServer if we are using VirtualBox machinery.

For other machineries this pull request adds more detailed error reporting if the error is related to  EADDRNOTAVAIL (cannot assign requested address). 

Handling of exceptions while initializing the ResultServer is now done in a separate function called 'handle_init_exceptions'.

More discussion about the feature can be found here: https://github.com/cuckoobox/cuckoo/issues/579

### DIFF
--- a/lib/cuckoo/core/resultserver.py
+++ b/lib/cuckoo/core/resultserver.py
@@ -9,8 +9,6 @@ import logging
 import datetime
 import SocketServer
 from threading import Event, Thread
-import subprocess
-import shlex
 
 from lib.cuckoo.common.config import Config
 from lib.cuckoo.common.constants import CUCKOO_ROOT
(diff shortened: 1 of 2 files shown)


jbremer:
Given this is specific to VirtualBox it would probably be more appropriate to handle such issues in `modules/machinery/virtualbox.py`. Yes, we're aware of this issue, and it's been bothering people for a while, but I think that with regards to the implementation we can do a slightly better effort.
Thanks for pushing in this direction though :)

lehmz:
Refactored the code to use default values from the config-files. Moved the interface initialization code to virtualbox.py.
 
After moving the code to virtualbox.py I realized that ResultServer() is called in cuckoo.py in cuckoo_init -function and machinery is initialized through the scheduler later on in cuckoo_main -function. 
This means that ResultServer will fail to bind to the non existing interface before anything in virtualbox.py is called.
I'll ask for advice on IRC on how to implement this correctly :)

PR #682, opened 2015-11-15
Exception handling for EADDRNOTAVAIL when starting the ResultServer

lehmz:
This PR just adds handling for EADDRNOTAVAIL (Cannot assign requested address) when starting the ResultServer. The exception handling will tell the user the most common reason (interface associated with the server is not up)  for the exception and provides a link to the documentation. 

https://github.com/cuckoobox/cuckoo/issues/579


jbremer:
I think this is much better, yeah :-) Will test it in a bit and merge it, thanks!

jbremer:
Thanks @lehmz! Sorry for taking such a long time to get back to this PR.

PR #683, opened 2015-11-15
Functionality to initialize the environment for individual machinery

lehmz:
This PR provides a generic way to initialize the environment for different machineries. It creates a new function 'initialize_environment' in the abstract class Machinery. That function is called from cuckoo.py cuckoo_init(). This gives us a way to initialize the environment (interfaces, etc) before we try to start ResultServer.   

For the third (and possibly the final) time this refers to the task/issue: https://github.com/cuckoobox/cuckoo/issues/579

### DIFF
--- a/cuckoo.py
+++ b/cuckoo.py
@@ -16,7 +16,6 @@ try:
     from lib.cuckoo.common.exceptions import CuckooDependencyError
     from lib.cuckoo.common.logo import logo
     from lib.cuckoo.common.utils import exception_message
-    from lib.cuckoo.core.resultserver import ResultServer
     from lib.cuckoo.core.scheduler import Scheduler
     from lib.cuckoo.core.startup import check_working_directory, check_configs
     from lib.cuckoo.core.startup import check_version, create_structure
@@ -77,7 +76,7 @@ def cuckoo_init(quiet=False, debug=False, artwork=False, test=False):
     if test:
         return
 
-    ResultServer()
+
 
     os.chdir(cur_path)
 
--- a/lib/cuckoo/common/abstracts.py
+++ b/lib/cuckoo/common/abstracts.py
@@ -83,10 +83,20 @@ class Machinery(object):
         """
         self.options = options
 
+    def _init_once(self):
+        """
+        Used to handle machinery specific initialization of the environment.
+        """
+        pass
+
     def initialize(self, module_name):
         """Read, load, and verify machines configuration.
         @param module_name: module name.
         """
+
+        #Initialize the environment if needed
+        self._init_once()
+
         # Load.
         self._initialize(module_name)
 
--- a/lib/cuckoo/core/scheduler.py
+++ b/lib/cuckoo/core/scheduler.py
@@ -567,6 +567,9 @@ class Scheduler(object):
         except CuckooMachineError as e:
             raise CuckooCriticalError("Error initializing machines: %s" % e)
 
+        #Starting the ResultServer
+        ResultServer()
+
         # At this point all the available machines should have been identified
         # and added to the list. If none were found, Cuckoo needs to abort the
         # execution.
--- a/modules/machinery/virtualbox.py
+++ b/modules/machinery/virtualbox.py
@@ -10,6 +10,7 @@ import logging
 import subprocess
 import os.path
 
+from lib.cuckoo.common.config import Config
 from lib.cuckoo.common.abstracts import Machinery
 from lib.cuckoo.common.exceptions import CuckooCriticalError
 from lib.cuckoo.common.exceptions import CuckooMachineError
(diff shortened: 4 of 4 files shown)


botherder:
This is an important addition to have to finally fix the interface issue. Unless @jbremer already fixed this, I would urge @jekil to look into ways to merge this.

jbremer:
I'd be in favor of renaming the function to something more simplistic, e.g., `init_once()`.
Also the implementation needs a bit of cleanup - among other things, creating a shell-like string only to split it again isn't really how one should craft command-line arguments, even if there are no security issues (as is the case here).

jbremer:
Furthermore some logic can be simplified due to various recent changes. A machine object will always have the `interface` and `resultserver_ip` set, so those try/except can be omitted.

lehmz:
I updated the PR and cleaned up the code a bit. Try-catch blocks have been removed and the command line argument is created in a less idiotic way :)

A machine object would have `interface` and `resultserver_ip` set, but from what I understood from the code is that the machines will be initialized when `sched.start()` is called from cuckoo.py after the ResultServer has already been started. If the initialization of the machine manager could be done earlier then I would have access to the initialized machines but I'm really not sure if that's possible.

jbremer:
Sorry, I meant to rename `initialize_environment` to `init()` or `init_once()` or something along those lines. Also, calling this from the initialization part of the `Scheduler` would make more sense than magically calling a function from the currently loaded machinery, I think. Anyway, this is just coding style preferences.

More interestingly. What happens if somebody accidentally copied the wrong IP address into the Cuckoo configuration. Will they now end up with a random `vboxnet` interface? What do we do when people use a different networking mechanism than expected (i.e., *not* a hostonly interface)?
I believe this would be a good addition for our upcoming `2.0-rc2` release, but some extra engineering efforts are required. @botherder @rep @jekil

lehmz:
I propose that we check the network configuration of the guest with VBoxManage showvminfo. We could initially just handle the case where the guest has exactly one interface configured and that is a hostonly interface. 

For more exotic configurations the users would need to understand the error message the ResultServer gives when it fails to bind to an interface. 

By looking at the code I think that the Scheduler will be started after the ResultServer has been started. ResultServer() is called in cuckoo_init and scheduler is started after that in cuckoo_main. So ResultServer will fail to bind to an interface before the scheduler has been started. I don't know the code base well enough to say if we could move starting the ResultServer to the Scheduler as well.

jbremer:
Fair enough - checking if there's just one network interface which is a hostonly interface should do the job.
You're right, I hadn't checked that yet. To de-duplicate the effort that you're currently doing perhaps we should move the `ResultServer()` initialization to *after* the following line: https://github.com/cuckoosandbox/cuckoo/blob/master/lib/cuckoo/core/scheduler.py#L564 and then include a call to a `self.init_once()` in the `machinery.initialize` method: https://github.com/cuckoosandbox/cuckoo/blob/master/lib/cuckoo/common/abstracts.py#L90-L94 which does the magic of creating the required network interfaces.

Does this make sense? :-)

lehmz:
It does make sense! :)

I updated the PR so that the ResultServer is now started from the Scheduler as you suggested. 

Added a additional clause that we try to bring up the interface only if the guest machine has exactly one interface configured and it's a hostonly interface.

jbremer:
I'll give it a little bit of code cleanup later on, but the functionality seems rock solid to me.
@botherder @jekil any thoughts on this PR?

jbremer:
So I'd prefer to see our existing functionality being reused a bit more, e.g., `self.machines()`. However, in order to be able to call `self.machines()` and have it return all the machines (as defined in the database now), we have to call `self._initialize()` which in turn requests the ResultServer port through `ResultServer().port`, introducing a circular dependency.
To resolve that issue we probably have to (partially) rewrite the ResultServer logic itself..

In other news, I was just testing this, and a crucial part is still missing. Before configuring the IP address one first has to create the hostonly interface; `vboxmanage hostonlyif create`. But we start spawning dozens of hostonly interfaces to users we should probably also check whether the interface already exists; `vboxmanage list hostonlyifs`.. so some stuff left to do.

Would you like to pick up on it?

jhg:
When this PR will be merged? I find a bug about vboxnet0 that when a analysis end this change to down and next analysis fail when try to bind. This PR also fix it?

jbremer:
I'm not sure about that. Are you using VirtualBox 5 by any chance?

lehmz:
I'll try to finish the PR later this week and add those checks to create the (correct) interfaces if they don't exist.

jbremer:
Great @lehmz!

lehmz:
Another week, another version. 
I'll try to sum up the intended functionality below. 

https://github.com/lehmz/cuckoo/blob/feature-vbox-interfaces/modules/machinery/virtualbox.py#L56-L64 
Check if the vboxnet0 interface exists. If it doesn't we don't do anything because we don't know what's the intended configuration. If the interface exists and VirtualBox sees the state as Up we don't do anything. So basically we only proceed if the interface exists but VirtualBox sees the state as Down

 https://github.com/lehmz/cuckoo/blob/feature-vbox-interfaces/modules/machinery/virtualbox.py#L67-L87
If even one of the configured guests has only a single interface configured and it's type is hostonly interface and the interface name is vboxnet0 we try to bring up the interface vboxnet0 with VBoxManage. 

I've tried to be really careful with the functionality so that we don't break any existing configurations. This PR will handle only the single case where the vboxnet0 interface is configured, assigned to a guest but isn't up when the ResultServer should be started.

jhg:
I'm using VirtualBox 5, yes

jbremer:
Any update on this matter @lehmz?

lehmz:
Nothing since the last PR I made on 20.3.2016 :)
I think that the PR addressed most of the issues when I made it but then again a lot has changed in the last five months or so.   

If the PR is still relevant or it has parts that can be used with the current master, please let me know. I have a ~full schedule for the next couple of weeks though.

jbremer:
Thanks again for the PR, but with recent updates from VirtualBox (I think, anyway) this doesn't appear to be required anymore. People, myself included, have been using VirtualBox 5.0 without issues etc. Closing PR.

PR #685, opened 2015-11-17
community.py fix & typo

Titotix:
25e3d4daaa481d772d0709673b4048d121fb41b3 : Complete fix from e756a3a6032b42d6294a24f596562af5c2050df8. Community script was not going to install anything if options -b 2.0 without the rewrite option (-w) was set.


jbremer:
Looks nice @Titotix, thanks. But how does `community.py` work without giving it the `-w` flag, what is it supposed to do? Off the top of my head I don't see the exact changes these commits push through.

Just tested without `-w` but with your changes and I'm getting a *lot* of `file skipped` messages, so doesn't look that good to me :-)

Titotix:
The `file skipped` are caused as the script go over the same files twice.
I will fix if you agree with me : for now `community.py` without `-w`, I assume that you already run this script once for branch 2.0, won't go in repository to check if there is new files.
So, for me the problem is:
1. You run the script once, it works.
2. Few time after, you want to update your signatures but without rewriting your current one (let's assume you did some local changes)
In this case the script will see that you already have the repo `network` (by example) and won't go over it to install any new files from github to it.

Even if I would not understand why, it could be an attending behavior ?

jbremer:
Although this PR doesn't merge at the moment, I guess these changes are fine?
If anyone has the time to test this, that'd be great. @botherder ?

jbremer:
I'm going to close this issue as the `community` integration has been rewritten for our upcoming `Cuckoo Package` work, see also #863. I believe these issues should be resolved in that version.
Took the liberty to update the documentation as was done in this PR, see also 01b1e827239026acc8f28fe53516f27b06c89d55.
Thanks again!

PR #691, opened 2015-12-01
update machinery default conf

Titotix:
All machinery default conf update to reflect recent changes


jbremer:
Actually more appropriate would be to define the networks as done in 3217495746ec5b82349fcfa794d52f3bc0f5a0ea (`default network interface`). In this PR you define a network interface specific to each VM (by putting it in the first machine definition) instead of making it a global default which is the preferred option, so to prevent people from unnecessarily defining the per-machine interface 20 times in a row. Agreed? :-)

Titotix:
Ya, agree, I missed this.

PR #692, opened 2015-12-01
Add API call to get analysis package's files

Titotix:
Add `/tasks/report/<task-id>/package_files` API call to be able to retrieve files uploaded to host by analysis packages.
The name `package_files` is not really relevant but I didn't find something else... `analysis`, `analysis_package`... ? Actually I just took the name of the folder where those files are uploaded. The choose of using an underscore is also to discuss.


jbremer:
Thanks! Looking good :-)

PR #693, opened 2015-12-02
doc additions

Titotix:
Just to add the value of key `status` in the json answer by api call `tasks/view`


jbremer:
Some people are too good for this world - keep it coming! :-)

PR #700, opened 2015-12-12
fix http 405 response when POST a search term

leibiao:
`@require_safe` will block the POST method and let search result be a http 405 response.So replace it with `@require_http_methods `


jbremer:
Thanks! Looking good.

PR #705, opened 2015-12-16
Update machinery interface conf

Titotix:
See closed pull request #691 
I took my motivation to do as @jbremer suggest =)


jbremer:
Good stuff, thanks!

Issue #702, opened 2015-12-14
JSON processtree/behavior PID mismatch


ssobr:
Looks like the parent pid is being listed as the pid in the behavior section. Here's an example:

"processtree": [
            {
                "pid": 1568, 
                "process_name": "invoice1211_pdf.exe", 
                "first_seen": 1449658424.688, 
                "ppid": 220, 
                "children": [
                    {
                        "pid": 240, 
                        "process_name": "dtayy.exe", 
                        "first_seen": 1449658169.029, 
                        "ppid": 1568, 
                    }
                ]
            }
"behavior": {
        "generic": [
            {
                "first_seen": 1449658424.688, 
                "process_name": "invoice1211_pdf.exe", 
                "pid": 220,
            }, 
            {
                "first_seen": 1449658169.029, 
                "process_name": "dtayy.exe", 
                "pid": 1568,

jbremer:
@rep care to take a look?

jbremer:
Thanks for resolving the issue.

PR #706, opened 2015-12-16
Fixed pid/ppid in JSON behavior

ssobr:
In regard issue #702, I think I found the typo. The pids appear to be correct now.


jbremer:
You're right, this seems to have been a typo by me :-) Thanks!

PR #707, opened 2015-12-18
Human module: button to click

Titotix:
Some targeted malware could use other languages than english for social-engineering.
So here is a translation in french of keywords listed to click on in human module and a few more as well.

### DIFF
--- a/analyzer/windows/modules/auxiliary/human.py
+++ b/analyzer/windows/modules/auxiliary/human.py
@@ -26,18 +26,18 @@ RESOLUTION = {
 def foreach_child(hwnd, lparam):
     # List of buttons labels to click.
     buttons = [
-        "yes",
+        "yes", "oui",
         "ok",
-        "accept",
-        "next",
-        "install",
+        "accept", "accepter",
+        "next", "suivant",
+        "install", "installer",
         "run",
-        "agree",
-        "enable",
-        "don't send",
-        "continue",
-        "unzip",
-        "open",
+        "agree", "j'accepte",
+        "enable", "activer",
+        "don't send", "ne pas envoyer",
+        "continue", "continuer",
+        "unzip", "dezip",
+        "open", "ouvrir"
     ]
 
     # List of buttons labels to not click.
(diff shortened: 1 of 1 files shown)


botherder:
It's a fair point, but we should discuss how to handle all languages then. @jbremer @rep @jekil thoughts?

jbremer:
I'm totally fine improving this on an as-needed basis. Ready to merge if you ask me. Perhaps if we have some more interns one day this could be a fun extra feature ;-)
Thoughts @botherder?

jekil:
I would suggest to have only constants defined in the aux module and a translation file (there are many formats created for i18n).

jbremer:
Well, I'm not so sure. It's not about what the dictionary says; it's about what the actual French installers require. I say we merge this PR and if people are interested to add other languages, they're free to do so.

That is, @Titotix, I assume that you've seen these words actually being used in French software? To the point where, with these translated words included, you managed to get past the whole installer / whatever?

jekil:
@jbremer I was not referring to dictionaries or whatever, and i don't want to add translations. Just to use a data file with some kind of structure where people can add values they found, better then add them to a list (after a while we will forgot what a word means).

jbremer:
Agreed, but in this case the translation is next to the English word, so I think that's fine, no? :-)

jekil:
Sure it is. In this case, my point isn't about this case.
It was about having a simple way people can add support for other languages and other "buttons" editing a txt file. 
I am ok for merging, my comment was about evaluating a more user friendly method to add values ;-)

jbremer:
Definitely agree. Based on the number of additions to this list I'm not sure if it's worth the effort, though.
If you feel like working on it, be my guest. Up to you :-)
Regardless of that todo item (for which a separate ticket may be appropriate), shall we merge this PR?

Titotix:
The french translation, yes I saw those title on legitimate or not software.
For the three that I added I just saw `save` but I thought it was not completely out of context to have `launch` and `execute`.
But I agree on the point of the discussion, in future it will need a proper way of handling these data out of the human module.
Thanks for the merge.

nahaye:
The spanish translation, someone have any clue about adding "Sí"  with an accent I put this.. but not working -*- coding: utf-8 -*-

Issue #15, opened 2011-12-16
Copyright Notice and Contributors


ghost:
I am annoyed by the gazillion count of the name of the author everywhere in the code. Just as a suggestion:
http://stackoverflow.com/questions/1497756/declaring-copyright-in-a-foss-project-with-major-and-minor-contributors

We as the people of the world are really thankful for any contribution i.e. to FOSS but it does not shed a good light if the authors (and there have already been more than one) put notes like:

"Cuckoo Sandbox is property of Claudio Guarnieri" (As seen on the html reports)

As FOSS is property of the community and not one entity. What the author(s) can do though is to dual license it to there liking but for this they have to get the consent of all the contributors as small there contribution might have been or not.

No offence Claudio we really appreciate the great work you have done but don't put a shadow on your work with this behaviour. Explicit is not always better then implicit.

PR #709, opened 2015-12-22
Better granularity on summary log from NtCreateFile/NtOpenFile API

Titotix:
/!\ This PR makes senses only with [PR #15 in monitor repo](https://github.com/cuckoobox/monitor/pull/15)
Summary log will be categorised according to status_info flag.
status_info is actually IoStatusBlock->Information arg from NtCreateFile/NtOpenFile
For details on IoStatusBlock arg, see [NtCreateFile doc](https://msdn.microsoft.com/en-us/library/bb432380%28v=vs.85%29.aspx)

So with this we actually know what happen precisely on files when NtCreateFile is called and probably we can know in which purpose the call has been used (just check file existence, open it, ...)

### DIFF
--- a/modules/processing/platform/windows.py
+++ b/modules/processing/platform/windows.py
@@ -197,7 +197,14 @@ class BehaviorReconstructor(object):
         self.files[arguments["file_handle"]] = arguments["filepath"]
         if NT_SUCCESS(return_value):
             status_info = flags["status_info"].lower()
-            return (status_info, arguments["filepath"])
+            if status_info in ("file_overwritten", "file_superseded"):
+                return ("file_recreated", arguments["filepath"])
+            elif status_info == "file_exists":
+                return ("file_opened", arguments["filepath"])
+            elif status_info == "file_does_not_exist":
+                return ("file_failed", arguments["filepath"])
+            else:
+                return ("file_opened", arguments["filepath"])
         else:
             return ("file_failed", arguments["filepath"])
 
--- a/modules/processing/platform/windows.py
+++ b/modules/processing/platform/windows.py
@@ -152,45 +152,45 @@ class BehaviorReconstructor(object):
         fn = getattr(self, "_api_%s" % event["api"], None)
         if fn is not None:
             if "flags" in event:
-                return fn(event["return_value"], event["arguments"], flags=event["flags"])
+                return fn(event["return_value"], event["arguments"], event["flags"])
             else:
-                return fn(event["return_value"], event["arguments"])
+                return fn(event["return_value"], event["arguments"], None)
 
     # Generic file & directory stuff.
 
-    def _api_CreateDirectoryW(self, return_value, arguments, flags=None):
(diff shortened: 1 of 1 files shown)


jbremer:
Nice @Titotix. As there are *six* possible values for the `status_info` flag I'd be in favour of further narrowing their representation down to their actual meaning.

E.g., a `status_info` of `file_exists` would result in a different meaning than what the `file_exists` summary key generally represents (namely, that a file exists). In this case I'd say that the `file_exists` `status_info` actually means that we should be reporting `file_opened` instead of `file_created`.

Would you be interested to pick up on this or should I? Furthermore, the added `flags` parameter is good and I don't think it requires the default value of `None`, so let's remove that.

Titotix:
It makes sense I think to remap in a narrow way those value. I did not test against a lot of sample but I have never seen some of those flags value like `file_does_not_exist` I even doubt that it really appears but I found nowhere to check this...

Just let's agree on the "remapping" : 
`file_exists` -> `file_opened`
`file_does_not_exit` -> `file_failed`
`file_overwritten` -> `file_recreated`
`file_superseded` -> `file_recreated`

For the two first, it is feeling (and also yours)
For the two others, I consider Microsoft documentation :

> Overwriting a file is semantically equivalent to a supersede operation, except for the following:

> - The caller must have write access to the file, rather than delete access. This implies that, if the file has already been opened by another thread, it opened the file with the FILE_SHARE_WRITE flag set in the input ShareAccess parameter.
> - The specified file attributes are logically ORed with those already on the file. This implies that, if the file has already been opened by another thread, a subsequent caller of NtCreateFile cannot disable existing FileAttributes flags but can enable additional flags for the same file. Note that this style of overwriting files is consistent with MS-DOS, Windows 3.1, and OS/2 operating systems.

The only difference which we could care would be the access on the file implied by the call. I assume that actually we don't : if a sample has delete access on a file it will probably also have write access on it.

At first I thought to remap `file_superseded` or `file_overwritten` to the other one, but I think it will be confusing: NtCreateFile and NtOpenfile don't write any contents to file, it is later call to NtWriteFile or whatever which will write contents.

So we keep only three values, `file_created`, `file_opened` and `file_recreated`.
By the way, perhaps `file_emptied` or `file_cleared` would fit better than `file_recreated` ?
@jbremer thoughts ?

jbremer:
I like `file_recreated`! Let's go for those three (`file_created`, `file_opened`, `file_recreated`) :-)
I take it you want to go for it? :)

Titotix:
Yep, gonna do it in the afternoon.

jbremer:
Looking good @Titotix! I believe the PR is ready for merging, right? :-)

Titotix:
Yes it is fine like that.
I am thinking about more improvements by the way : NtCreateFile permits also to create and traverse directory, so it would be nice to log it properly as well. But I won't do it now. Should I open a issue ticket ?

jbremer:
Let's open a separate ticket for that :-) That way I can merge this one right now.
Great work @Titotix!

PR #716, opened 2015-12-24
create all.yar

liebesu:
more than 3000 yara rules I collected


botherder:
That's something that I would probably rather have on the community repository. @jbremer @jekil @rep thoughts?

jbremer:
I agree with @botherder. Also somebody will have to go through all those Yara rules to filter out any *bad* ones, depending on where they're coming from, @liebesu?

botherder:
True. If the rules aren't done well, the Yara scanning can take a serious amount of time and delay the reporting procedure.

jbremer:
I'm all up for adding more Yara rules to our repository, but as @botherder said it would make sense to move this to the Community repository. In addition to that I'm not too confident about 3000 semi-random rules.
As nobody stepped up (so far anyway) to make a selection of *proper* Yara rules I'm thinking of closing this issue - at least until somebody is going to do some research on all these rules. Does anyone have any results of running all these rules on a set of millions of samples?

jbremer:
Personally I don't think we're ever going to merge this. If somebody takes the time to properly cherry-pick Yara rules that would be a totally different matter, though.
Closing this issue.

PR #718, opened 2015-12-28
Update mongodb.py

xvsfekcn:
shots_path is had been set when current.set_path() ,so it's not need  get the shots_path again
and When I use a diffrent path in analyses path ,only need modify once


jbremer:
Would be even better to replace all instances of `shots_path` by `self.shots_path`, would you like to do or prefer if I pick up on that?

jbremer:
What I mentioned earlier didn't apply anymore. Thanks for the patch @xvsfekcn.

PR #719, opened 2015-12-28
Update resultserver.py

xvsfekcn:
when run more than 25 vms ,cuckoo will be block，and the code cuckoo block also is select() function,so i think it will be so many request to handle,which due to the file that opened large than the max fd number and when use the epoll ,i had never miss this problem again

### DIFF
--- a/lib/cuckoo/core/resultserver.py
+++ b/lib/cuckoo/core/resultserver.py
@@ -140,9 +140,13 @@ class ResultHandler(SocketServer.BaseRequestHandler):
         while True:
             if self.end_request.isSet():
                 return False
-            rs, _, _ = select.select([self.request], [], [], 1)
-            if rs:
-                return True
+            #rs, _, _ = select.select([self.request], [], [], 1)
+            #if rs:
+            #   return True
+            events = epoll.poll(1)
+            for fd,event in events:
+                if event & select.EPOLLIN:
+                    return True
 
     def seek(self, pos):
         pass
--- a/lib/cuckoo/core/resultserver.py
+++ b/lib/cuckoo/core/resultserver.py
@@ -137,6 +137,8 @@ class ResultHandler(SocketServer.BaseRequestHandler):
             self.rawlogfd.close()
 
     def wait_sock_or_end(self):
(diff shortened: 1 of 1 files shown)


jbremer:
I run the same amount of VMs regularly, but never noticed this exact issue. Aside from this workaround, have you tried upping the max fd limit?
@rep: Any thoughts on the PR itself?

xvsfekcn:
yeah,this is not a high probability problem,but i really miss this so many times when use 28 vms and run the sample which get it from vt and most of this samples are malware.I know that in include/linux/posix_types.h the __FD_SETSIZE   is 1024 ,so when fd is large than 1024 will be block,it need recomplie the kernel when want to modify this value

jbremer:
I'm running 30 VMs just fine usually, without requiring this patch. The file descriptor limit can be bumped (probably requires you to logout and login though).
Another fair point raised by @rep for not merging this PR is that `epoll` is not supported under Windows, thus rendering Windows support (as host) non-functional.

Unless anyone has an objection I'm going to close this issue without merging @rep @jekil @botherder.

jbremer:
Ah, right, I missed the real issue that you were referring too. I'm thinking of opting in on `poll(2)` support when available. This should allow platforms where one may run 30 VMs next to each other - usually Linux-based setups - with support for having socket fd's higher than 1024, but still remain Windows support through `select(2)`.

In the way that we use sockets here - one thread per socket - I'm not sure if we really require more semi-advanced `epoll(2)` support. It would definitely make sense to refactor this thing at some point to reduce the CPU overhead, but that's for another day; as-is it's working just fine.

jbremer:
Could you please try the commit above @xvsfekcn? :-) It uses `poll(2)` instead of `select(2)` on platforms where that functionality is available.
Now looking at it I don't see a lot of difference between `poll(2)` and `epoll`, but oh well, this should do.

xvsfekcn:
Yeah,i really only think about it's run in linux.I will try your solution in the code running in the centos6.5.but I think epoll is better than poll,because epoll is the most efficient in linux about i/o
mechanism

PR #724, opened 2016-01-06
Create XEN and LibVMI document as restructured text

NiickB:
XEN and LibVMI documentation in restructured text


jbremer:
Good start, but not to be merged yet.

PR #734, opened 2016-01-21
Avoid usage of hard-coded URLs in AJAX calls

tsmolka:
URLs in JavaScript might not be correct if for example user sets FORCE_SCRIPT_NAME and Cuckoo is deployed in a subdirectory (e.g. http://localhost/cuckoo/).


jbremer:
I'll take it, thanks!

Issue #746, opened 2016-02-02
LINUX platform wget error


aspel:
Cuckoo Sandbox 2.0-rc1
 www.cuckoosandbox.org
 Copyright (c) 2010-2015

 Checking for updates...
 Good! You have the latest version available.

2016-02-02 17:38:20,404 [lib.cuckoo.core.scheduler] INFO: Using "qemu" as machine manager
2016-02-02 17:38:20,927 [lib.cuckoo.core.scheduler] INFO: Loaded 2 machine/s
2016-02-02 17:38:20,938 [lib.cuckoo.core.scheduler] WARNING: As you've configured Cuckoo to execute parallel analyses, we recommend you to switch to a MySQL ora PostgreSQL database as SQLite might cause some issues.
2016-02-02 17:38:20,959 [lib.cuckoo.core.scheduler] INFO: Waiting for analysis tasks.
2016-02-02 17:52:50,692 [lib.cuckoo.core.scheduler] INFO: Starting analysis of URL "http://xxxx/Hd4D6v72AQREm7JlD" (task #7, options "route=none")
2016-02-02 17:52:50,814 [lib.cuckoo.core.scheduler] INFO: Task #7: acquired machine vm1 (label=vm1)
2016-02-02 17:52:50,823 [modules.auxiliary.sniffer] INFO: Started sniffer with PID 8975 (interface=eth0, host=192.168.0.11, pcap=/opt/cuckoo/storage/analyses/7/dump.pcap)
dropped privs to root
tcpdump: listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
2016-02-02 17:52:51,841 [lib.cuckoo.core.guest] INFO: Starting analysis on guest (id=vm1, ip=192.168.0.11)
2016-02-02 17:58:10,767 [lib.cuckoo.core.scheduler] ERROR: Analysis failed: Unable to import package "modules.packages.wget", does not exist.
0 packets captured
2 packets received by filter
0 packets dropped by kernel

2016-02-02 17:52:50,032 [root] DEBUG: Starting analyzer from: /wvcbmjdypq
2016-02-02 17:52:50,041 [root] DEBUG: Storing results at: /tmp/DDKRessJSC
2016-02-02 17:52:50,053 [root] DEBUG: No analysis package specified, trying to detect it automagically.
2016-02-02 17:52:50,058 [root] INFO: Automatically selected analysis package "wget"
[...]

PR #761, opened 2016-02-12
created a new linux module "wget"

aspel:
I created a new linux module "wget"  issue #746
Wget is a module to direct download a malware and execute it.


jbremer:
Nice! Perhaps an idea to randomize the name, though? Although I'm not sure what would be appropriate..

jbremer:
I'm going to close this issue as we sort of decided that having Cuckoo do `wget` doesn't really make too much sense. So the Linux Analyzer should probably be updated to reflect that decision. Thanks though.

PR #778, opened 2016-02-22
WinAPI processing: forgotten file_created status_info #709

Titotix:
Just come to realize that I forgot one the value, `file_created`


jbremer:
Thanks :)

PR #779, opened 2016-02-22
Fix memory list issue

mkmanwich:
Jsonify doesn't work on plain arrays, without a key.
Getting 500 internal server error without this fix.


mkmanwich:
API memory/list bugfix

jbremer:
Thanks! Seems I introduced this bug :-)

PR #787, opened 2016-02-27
decoding to latin1 doesn't work for me, but this does

razuz:
For some reason decoding to latin1 didn't make any difference, but when the string is encoded to UTF-8 then works. The string itself will look like garbage still, but at least it doesn't crash the process.
For example :
2016-02-27 15:32:14,750 [analyzer] INFO: Injected into process with pid 3044 and name 텀ȭ


jbremer:
Thanks @razuz! :)

PR #789, opened 2016-02-28
Fix crash when suricata files dir doesn't exist

razuz:
When suricata files dir doesn't exist (for example files-store is not enabled) then this leads to crash : 
`
suricata error : 
  File "/srv/cuckoo/lib/cuckoo/core/plugins.py", line 215, in process
    data = current.run()
  File "/srv/cuckoo/modules/processing/suricata.py", line 286, in run
    self.parse_files()
  File "/srv/cuckoo/modules/processing/suricata.py", line 209, in parse_files
    for filename in os.listdir(files_dir):
OSError: [Errno 2] No such file or directory: '/srv/cuckoo/storage/analyses/475/suricata/files'
`

### DIFF
--- a/modules/processing/suricata.py
+++ b/modules/processing/suricata.py
@@ -206,6 +206,14 @@ class Suricata(Processing):
 
         # Index all the available files.
         files_dir = os.path.join(self.suricata_path, self.files_dir)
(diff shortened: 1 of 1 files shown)


jbremer:
Excellent, @razuz. Thanks!

PR #797, opened 2016-03-07
Fixes #796, if Machinery module raises an exception, unroute_network(…

### LINKED ISSUE #796: unroute_table crashes when machine startup is broken
If Machinery.start() raises an exception, we don't get to execute code on route_network, and sice interface and rt_table are not defined on __init__, it fails nongracefully.

This can be easily fixed by putting self.interface = None and self.rt_table = None in __init__ on AnalysisManager class

XayOn:
…) will crash


jbremer:
I fixed this at the time before seeing your PR - thanks again.

XayOn:
Heh, don't worry, I usually accompany tickets with PRs if I can.

PR #803, opened 2016-03-09
Drop Suricata privileges to enable Cuckoo to access them 

nsmfoo:
As Suricata in socket mode is being run as root, files created is only accessible by root, Which makes Cuckoo not being able to read the content of the files create by Suricata.

This fix includes a description of what you need to change in the Suricata configuration file in order to make Suricata drop privileges to the set Cuckoo user.

The fix also enables the non root user to create the socket file in /var/run/suricata which is something that requires root privileges.


jbremer:
Thanks! Good stuff :)

PR #805, opened 2016-03-10
Added linux malware analysis prototype.

0x71:
In my bachelor thesis I developed a prototype that can be used for comprehensive **static and dynamic Linux malware analysis**. As this prototype is based on the Cuckoo Sandbox, it is used to automatically run and analyze files inside an isolated Linux operating system and collect several analysis results that outline the malware bevavior. Furthermore, since I adapted and expanded the source code of Cuckoo Sandbox the prototype is capable of running both Linux and Windows malware analysis tasks.

#### What I've done so far:
* Development of modules that run dynamic analyses:
  * SyscallTracer: Run, control and observe malware
  * FilesystemTracer: Observe filesystem activities using kernel modules
  * ResultLogger: Collect analysis results and directly send them to the host (For this, I adapted the reporting and communication protocols that were already implemented for Windows malware analysis)
* Expansion of existing modules that run static analyses:
  * Implemented static analysis of ELF files
* Several smaller changes:
  * Webinterface (the user can choose between linux and windows analysis)
  * Reporting module (formatting and output things)
  * Linux analyzer module (to make my modules run and communicate properly)
  * Created a netlog module (according to cuckoomon) for host-guest result transmission using a linux guest 

* Code update to the latest version of Cuckoo Sandbox


jbremer:
Hi Nils,

Sounds like some great work that you did here. Two things though:

* Since some time we have Linux analysis support by @rep. If your solution uses a different approach then I'd be more than happy to somehow look into merging the two approaches or seeing how we could have both possibilities in parallel.
* Due to merging you've included way too much of non-interesting stuff to this PR. Would you be so kind to attempt to provide us with the same merge, but without all the superfluous changes? :-) One way to do this could be to rebase all of your changes on top of the latest and greatest `master` branch. You'd loose some information, but as-is we can't work with the PR.

Thanks!

Jurriaan

0x71:
Hi Jurriaan,

Thanks for your feedback!

I will provide you the same merge, but without all the small changes ;-) (as soon as possible). 

At first sight, our solutions use slightly different approaches. While your currently implemented solution seems to use stap and strace in order to observe system-wide syscalls, my solutions utilizes the ptrace syscall to monitor every syscall made by the malware and all its subprocesses.      Therefore, the combination could work well. We need to follow up this idea.

Best regards,
Nils

jbremer:
Thanks! Sounds good. Arguably `ptrace` is similar to `strace`, but your approach would surely allow much more flexibility.

Jurriaan

botherder:
@jbremer, before anything of this size is merged into upstream, it shall be discussed with the rest of the team.

jbremer:
Naturally, @botherder. Closing this issue as the PR has been improved a lot in #807.

PR #806, opened 2016-03-10
bugfix datetime serializer in jsondump.py

knifeyspoony:
The 'default' serializer in jsondump.py was adding microseconds/1000.0 to the timestamp. However, in order to get the milliseconds that are dropped by timetuple(), this should have been microseconds/1000000.0

See http://stackoverflow.com/a/26161653.

Cheers

### DIFF
--- a/modules/reporting/jsondump.py
+++ b/modules/reporting/jsondump.py
@@ -16,7 +16,7 @@ def default(obj):
     if isinstance(obj, datetime.datetime):
         if obj.utcoffset() is not None:
             obj = obj - obj.utcoffset()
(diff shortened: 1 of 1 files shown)


jbremer:
Wow, amazing.. thanks a lot! Actually we've discussed not too long ago to get rid of the absolute timestamp in favor of a simple `microsecond since process start` relative time, but that will be pushed through later on. In the meantime I'm happy to take this improvement.

Thanks again!

jbremer:
Actually, hehe, I believe you're missing a zero.. so adding that :-)

PR #819, opened 2016-03-19
IRC question

rvargasbarranco:
Some months ago I asked how to run the distributed script, I´ve just checked that the new information related with the dist.scheduler and dist.status was added, but I also had problems with the DB, for this reason i think my changes could be a good idea in the documentation.


jbremer:
I don't have much to add on this :) Thanks.

PR #820, opened 2016-03-20
Checking the distributed in a lab

rvargasbarranco:
Related with my master thesis, I did some experiments with the distributed script, so I´ve found some errors in the documentation.

1) Wrong use of the API to register the nodes.(/api/node)
2) It´s mandatory to use one instance script per node, as we talked some month ago on IRC.

Also I think that could be interesting more information related with the threshold parameter, I tested my lab with to nodes (2 virtual machines per node), and I needed a value around 2 to see some task in the secondary node.

I hope this information could be interesting for the project.

Kind regards


jbremer:
Thanks!

Issue #818, opened 2016-03-18
Invalid binary symlink causes /tasks/report/<id>/all to fail


knifeyspoony:
When cuckoo.conf specifies delete_bin_copy, scheduler.py will delete the copy upon completion. On *nix this will invalidate the symlink in the analysis directory. When the report for this analysis is requested through the api /tasks/report/id/all, tarfile's add function will raise an OSError when it hits the invalid symlink. 

Two proposed solutions are:
1. In scheduler.py if the delete_bin_copy flag is present, check to see if you're left with an invalid symlink after deleting self.binary, and os.unlink() it if so.
2. Just swallow the OSError raised by tar.add() in the /tasks/report/id api calls.

I would have just issued a pull request, but didn't know which solution (if any) was preferred!

Cheers

jbremer:
We should probably implement both methods. Delete the symlink if the binary copy is removed (through the options) and make sure the file still exists before adding it to the tar.
Would you like to try to implement this or should I?
Thanks for reporting!

knifeyspoony:
I can give it a go. I'll send up a pull request asap. No problem!

jbremer:
Merged the PR, thanks!

PR #821, opened 2016-03-21
Handle invalid symlinks when delete_bin_copy is specified.

knifeyspoony:
This addresses issue #818 

When delete_bin_copy is specified, the binary in the report directory is checked to see if it is now an invalid symlink, and it's deleted if necessary. In the API, report files are checked for existence before being added to the tar (kind of double-checking here, since we are making sure to delete the invalid symlink).


jbremer:
Looking good! Will merge it after testing it quickly later. Thanks!

jbremer:
Works as advertised - thanks @knifeyspoony!

PR #832, opened 2016-03-30
generic package bug fix

doomedraven:
missed value for title, if title arg not specified, it tries to interpret rand_title as file, in this way works fine :)


jbremer:
@jekil @botherder? :)

doomedraven:
here only one problem, all args must be passed before title, 
like this

```
args = ["/c", "start", "title", rand_title, "/wait", path]
```
if not, all values after title args interpreted as title part even if is other arg

and now it works correctly

PR #839, opened 2016-04-04
Fixed the URL

athend:
The URL was http://www.cuckoofoundation.org/contribute instead of http://www.cuckoofoundation.org/contribute.html hence it was giving 404.


jbremer:
Thanks!

PR #850, opened 2016-04-08
Return a jsonified object when the report format is json

mwielgoszewski:
When calling `GET /tasks/report` [1], the API does not return the proper content-type. This PR changes the content-type returned to `application/json`, rather than the default `text/html; charset=utf-8`.

[1] http://docs.cuckoosandbox.org/en/latest/usage/api/#tasks-report


jbremer:
You're correct, thanks!

PR #851, opened 2016-04-08
Windows cmd title needs to be a quoted string

mwielgoszewski:
The title needs to be a quoted string. A breaking change was introduced in 4a90e0a893407569b42c417bb5c3b4c00f092cac when moving to list based arguments.

### DIFF
--- a/analyzer/windows/modules/packages/generic.py
+++ b/analyzer/windows/modules/packages/generic.py
@@ -18,7 +18,7 @@ class Generic(Package):
         cmd_path = self.get_path("cmd.exe")
 
         # Create random cmd.exe window title.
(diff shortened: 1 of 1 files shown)


doomedraven:
but it also missed "title" before fake title name

mwielgoszewski:
@doomedraven ? The command syntax for `start` does not call for putting "title" before the title argument.

doomedraven:
See this merge request https://github.com/cuckoosandbox/cuckoo/pull/832

In my all vms if title not specified, start interpreting fake_title as file to start

jbremer:
Thanks, in the meantime this has already been taken care of, though. Guess I missed this PR. So closing this PR :-)

PR #855, opened 2016-04-12
File type for JAR

rk700:
The python magic module returns "Java archive data" for JAR file.

Tried on Mac and Linux(Debian, Archlinux).


jbremer:
Thanks!

PR #863, opened 2016-04-18
Turn Cuckoo into a Python package

jbremer:
This PR introduces a pretty major rewrite of the way that Cuckoo is structured and operated. It aims at improving the user experience through the following improvements - improvements that we should have addressed a while ago.
- Users will now be able to `pip install cuckoo` in order to install the latest version of Cuckoo Sandbox.
- By uploading Cuckoo as Python package we'll be able to push updates more frequently (e.g., bug fixes reported by more than 1 or 2 users).
- Being a package we control which packages are installed at the system of a user - at a version that we know is supported - making setup easier for users and bug reports more accurate for us.
- The configuration is no longer part of the Git repository. This makes it possible to update Cuckoo without affecting a users configuration. In updates that actually modify the configuration Cuckoo will have to point out to users what has changed, how they can update it, and verify that they have done so correctly. Hopefully this will lead to more organized configuration management (e.g., field A should be an integer, field B should be a path, etc).
- We introduce the `Cuckoo Working Directory`, the directory where all user configuration, configurable Python components (i.e., Signatures, Analyzer, etc), Yara rules, Cuckoo storage, etc is saved.
- With the `CWD` (`Cuckoo Working Directory`) it is now possible to maintain multiple Cuckoo instances on a machine without having to install Cuckoo multiple times, simply by running Cuckoo under a different `CWD`.
- The `CWD` (which defaults to `~/.cuckoo`) is automatically created the first time Cuckoo is run. A user can then modify his or her configuration, pull the latest Signatures from the Cuckoo Community, etc, and after that run Cuckoo.
- The concept of `Cuckoo Apps` is introduced - no longer are scripts from the `utils/` directory required. I'll go more in-depth on this later in the new Cuckoo workflow.
- Making installing a Cuckoo update as easy as `pip install -U cuckoo`.
# Cuckoo Workflow

With all these changes the way users will operate Cuckoo will also change - in a good way. I mentioned `Cuckoo Apps`, these will be used to invoke the various components of Cuckoo. First of all let's install and setup a new Cuckoo instance. We'll be using a `virtualenv` here which, although it is not required, it is recommended. Note that `(cuckoo)$` simply shows that you're inside the `virtualenv` named `cuckoo`.

``` bash
$ virtualenv cuckoo
$ . cuckoo/bin/activate
(cuckoo)$ pip install cuckoo
(cuckoo)$ cuckoo
... [snip] ...
```

Cuckoo will now be installed (in the `virtualenv`) and the default configuration will be setup in `~/.cuckoo`.  Generally speaking a user will want to modify some configuration, e.g., the name of the virtual machine. Exactly like before this can be done in the `conf/virtualbox.conf` file, the only difference is that this file is now located at `~/.cuckoo/conf/virtualbox.conf`.

It is recommended to install the 300+ Community Signatures. These aid in identifying interesting and malicious behavior. Before the Cuckoo package one would fetch these Signatures by running `./utils/community.py -waf`. This has now been transformed in a `Cuckoo App` and can be done as follows.

``` bash
# After running this command the Signatures will be installed.
(cuckoo)$ cuckoo community

# Show the help message of the cuckoo community app.
(cuckoo)$ cuckoo community --help
Usage: cuckoo community [OPTIONS]

  Utility to fetch supplies from the Cuckoo Community.

Options:
  -f, --force              Overwrite existing files
  -b, --branch TEXT        Specify a different community branch rather than
                           master
  --file, --filepath PATH  Specify a local copy of a community .tar.gz file
  --help                   Show this message and exit.

```

If you have a local copy of the community `.tar.gz` file (e.g., after running `wget https://github.com/cuckoosandbox/community/archive/master.tar.gz`) this can be imported as follows.

``` bash
(cuckoo)$ cuckoo community --file master.tar.gz
```

Now we're good to go let's submit some samples and URLs. Note that multiple tasks can be submitted at once.

``` bash
(cuckoo)$ cuckoo submit sample1.exe sample2.exe sample3.exe
Success: File "/home/jbr/sample1.exe" added as task with ID #1
Success: File "/home/jbr/sample2.exe" added as task with ID #2
Success: File "/home/jbr/sample3.exe" added as task with ID #3
(cuckoo)$ cuckoo submit --url google.com bing.com
Success: URL "google.com" added as task with ID #4
Success: URL "bing.com" added as task with ID #5
```

In order to actually analyze the tasks we'll have to start the Cuckoo daemon (as we used to do with `./cuckoo.py -d`). Not much changed to this command - it can now be invoked as follows.

``` bash
(cuckoo)$ cuckoo -d
```

This command will run indefinitely (unless a `maximum analysis count` was provided with the `--maxcount` switch, e.g., `-m 5`).

By default Cuckoo will process the raw analysis data into a Cuckoo report, however, if your Cuckoo instance runs multiple VMs in parallel it is smart to offload this CPU intensive process to one or more processing processes (because of the way Python works with the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock) etc). Before one would do this with, e.g., `./utils/process.py -p 4` or `./utils/process2.py p0`. The `process.py` utility has had unresolved issues for years with its `multiprocessing` component leaving it unusable (it would simply get stuck after a while) and `process2.py` was only compatible with `PostgreSQL`-based database backends. These two commands have been merged together into a simple Cuckoo App and can be invoked as follows.

``` bash
(cuckoo)$ cuckoo process instance1
```

A report can be re-generated, much like before, using the `-r` switch (e.g., `cuckoo process -r 1`). Note that in _daemon_ mode an `instance` name should be provided - this allows multiple `cuckoo process` instances to work next to each other. In other words, if one `cuckoo process` isn't able to keep up with the workload, you start another (e.g., `cuckoo process instance2`).

Having analyzed and processed the task it is now time to actually view it. Surely you'll be able to guess which command we'll be running in order to do this. Following two commands start the `Cuckoo API` and `Cuckoo Web Interface` respectively. These are the development HTTP servers though - putting them behind, say, `nginx` is still on the todo list (but is not that difficult). Note that both these processes run indefinitely so a separate terminal / screen session / background job is required.

``` bash
(cuckoo)$ cuckoo api 
 * Running on http://localhost:8090/ (Press CTRL+C to quit)
```

``` bash
(cuckoo)$ cuckoo web runserver
Performing system checks...

System check identified no issues (0 silenced).
April 18, 2016 - 15:22:26
Django version 1.8.4, using settings 'web.settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
```

Well, that's pretty much most of it already. There are a couple more goodies - please see as follows - but other than that this should be all to get started. We made sure that the Signatures and other features from the Cuckoo Community remain backwards compatible (even though [lib.cuckoo.common.abstracts](https://github.com/cuckoosandbox/community/blob/master/modules/signatures/windows/allocates_rwx.py#L5) is no longer a thing given that everything falls under the `cuckoo` package now, instead of `lib.cuckoo`, yay Python magic :P).

The remaining Cuckoo goodies are as follows.

``` bash
# Running the rooter.
(cuckoo)$ cuckoo rooter -v

# Cleaning the database(s), log files, and Cuckoo storage (i.e., `storage/analyses` etc).
(cuckoo)$ cuckoo clean
```

That's all for now! Up to an improved Cuckoo Workflow.

Regards,
Jurriaan


gtback:
This is exciting! One quick question: Is the idea that this would get merged in before the 2.0 final release, or will this be the foundation of "Cuckoo 3.0"?

jbremer:
@gtback I'd say 2.0 RC2 :-) No need to keep this laying around for what could easily be another year ;-)

swackhamer:
Any plans on merging this to master soon?

jbremer:
@swackhamer There are still quite some things that we'd like to address before the first release. Due to some new developments there are also some bugs that we'll have to get rid of first. Aside from that it's mostly stable ;-)

lovina37:
@jbremer will it work for android analysis??

jbremer:
@lovina37 nothing much changed regarding android analysis as we're currently not really focussed on that (as you know, probably). I'll see if my new guy can take a proper look at Android, but that will be somewhere in ~6 weeks from now probably.

PR #865, opened 2016-04-19
Android: Simple fix for race condition where agent is requested before it is ready

RuneTM:
The python agent that is pushed to the Android device is requested before it is ready, making it crash and the analysis fail. By inserting a short "sleep", the agent has time to become ready.

Here is an example of the output from cuckoo.py, when the agent is not ready:
2016-04-19 08:27:00,571 [modules.machinery.avd] INFO: Stopping AVD listening on port 5554
2016-04-19 08:27:01,735 [lib.cuckoo.core.scheduler] ERROR: Failure in AnalysisManager.run
  File "/home/rune/devcuckoo/cleanclone/lib/cuckoo/core/scheduler.py", line 482, in run
    self.launch_analysis()
  File "/home/rune/devcuckoo/cleanclone/lib/cuckoo/core/scheduler.py", line 374, in launch_analysis
    self.guest_manage(options)
  File "/home/rune/devcuckoo/cleanclone/lib/cuckoo/core/scheduler.py", line 290, in guest_manage
    guest.start_analysis(options, monitor)
  File "/home/rune/devcuckoo/cleanclone/lib/cuckoo/core/guest.py", line 382, in start_analysis
    r = self.get("/")
  File "/home/rune/devcuckoo/cleanclone/lib/cuckoo/core/guest.py", line 258, in get
    return requests.get(url, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 67, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/api.py", line 53, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 468, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/adapters.py", line 426, in send
    raise ConnectionError(err, request=request)
ConnectionError: ('Connection aborted.', error(104, 'Connection reset by peer'))

### DIFF
--- a/conf/auxiliary.conf
+++ b/conf/auxiliary.conf
@@ -42,10 +42,6 @@ script = data/mitm.py
 # be set to bin/cert.p12.
 certificate = bin/cert.p12
 
-# Path to the PEM formatted public certificate, somewhat like the above,
-# copy ~/.mitmproxy/mitmproxy-ca-cert.pem to analyzer/windows/bin/cert.pem
-public_cert = bin/cert.pem
-
 [services]
 # Provide extra services accessible through the network of the analysis VM
 # provided in separate, standalone, Virtual Machines [yes/no].
--- a/conf/auxiliary.conf
+++ b/conf/auxiliary.conf
@@ -42,6 +42,10 @@ script = data/mitm.py
 # be set to bin/cert.p12.
 certificate = bin/cert.p12
 
+# Path to the PEM formatted public certificate, somewhat like the above,
+# copy ~/.mitmproxy/mitmproxy-ca-cert.pem to analyzer/windows/bin/cert.pem
+public_cert = bin/cert.pem
+
 [services]
 # Provide extra services accessible through the network of the analysis VM
 # provided in separate, standalone, Virtual Machines [yes/no].
--- a/modules/auxiliary/mitm.py
+++ b/modules/auxiliary/mitm.py
@@ -5,7 +5,6 @@
 
 import logging
 import os.path
-import shlex
 import subprocess
 import threading
 
@@ -26,7 +25,6 @@ class MITM(Auxiliary):
         port_base = int(self.options.get("port_base", 50000))
         script = self.options.get("script", "data/mitm.py")
         certificate = self.options.get("certificate", "bin/cert.p12")
-        public_certificate = self.options.get("public_cert", "bin/cert.pem")
 
         outpath = os.path.join(CUCKOO_ROOT, "storage", "analyses",
                                "%d" % self.task.id, "dump.mitm")
@@ -48,13 +46,6 @@ class MITM(Auxiliary):
                       "aborted.", certificate, cert_path)
             return
 
-        cert_path = os.path.join("analyzer", "windows", public_certificate)
-        if not os.path.exists(cert_path):
-            log.error("Mitmdump public certificate not found at path \"%s\" "
-                      "(real path \"%s\"), man in the middle interception "
-                      "aborted.", certificate, cert_path)
-            return
-
         PORT_LOCK.acquire()
 
         for port in xrange(port_base, port_base + 512):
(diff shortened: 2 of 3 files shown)


jbremer:
At that time the Android emulator should be up-and-running, though. It can only get to the `r.get("/")` call (https://github.com/cuckoosandbox/cuckoo/blob/master/lib/cuckoo/core/guest.py#L382) if it has returned from `self.wait_available()` (https://github.com/cuckoosandbox/cuckoo/blob/master/lib/cuckoo/core/guest.py#L373) which pretty much waits until a listening socket is open at the other side (https://github.com/cuckoosandbox/cuckoo/blob/master/lib/cuckoo/core/guest.py#L269-L272).
Unless something weird is happening to your `guest status` (see also the `== "starting"` part), but that sounds unlikely.

RuneTM:
If you want to try and reproduce the problem, here's a relatively brief version of what I do when i setup a new cuckoo, it should be the same with 2.0 RC1 or the latest from repo. Note I do this from within a VM.
------------------------------------------------------------
Setup guide for cuckoo 2.0 RC1 for android use

# Get community data (python agent for android)
./utils/community.py -waf

# Edit configuration file auxiliary.conf
disable sniffer (emulator does this automatically)

# Edit configuration file conf/avd.conf 
Set emulator_path, adb_path and  avd_path
For the two first, do something like `which emulator` and `which adb`.
For the latter, you should probably just change the username to the user who will be running the emulator.

# Edit configuration file conf/cuckoo.conf
set machinery to avd
under the [resultserver] section, set ip to 127.0.0.1

# Edit configuration file processing.conf
Enable apkinfo and droidmon

# Edit configuration file reporting.conf
Enable reporthtml, mongodb

#Create android avd
# Start Android Virtual Device Manager
android avd

# Create the AVD
# AVD Name: cuckoo-bird
# Device: Nexus One
# Target: Android 4.1.2 (if you do not have it, you need to download the ARM system image for this version)
# CPU: ARM
# Skin: no skin
# Internal storage: 512 MiB
# SD Card size: 512 MiB
# Tick "Use Host GPU"

%% Start the AVD from the manager %%

# Install extra components
When the emulator device has started, execute the setup device script:
cd data/android/
./create_guest_avd.sh
Note: You can ignore the "rm failed" messages

# Last few configurations in emulator
Settings > Security > Screen lock > None
Settings > Display > Sleep > After 30 Minutes
Execute generate contacts app
Execute superuser app
Execute Xposed Installer
-> Press Modules -> Place tick at Android Blue Pill and Droidmon
-> Open Framework from dropdown -> Install/update -> Allow Superuser request -> Cancel reboot -> Soft reboot
When device has rebooted close emulator (with the "X" button, poweroff does not work for me)

# Start web server
cd web
python manage runserver 0.0.0.0:5000

# Start cuckoo
python cuckoo.py

snemes:
@jbremer, this is a glitch with how ADB port forwarding works. The Android machinery module sets up port forwarding between the host and the Android emulator via the 'adb forward' command. After this step however, the host port will be open and ready to accept connections, even if the remote port (in the emulator) is not in a listening state. Due to this, the code that tries to connect to the agent port on the host in order to check if it is available (https://github.com/cuckoosandbox/cuckoo/blob/master/lib/cuckoo/core/guest.py#L287) will not be reliable for the Android agent.

In my opinion this is a good pull request, and to be honest, this is exactly the same method I used to fix it, except that I used sleep(5) instead of sleep(10). :)

jbremer:
@snemes Oke, thanks for the feedback :-)

PR #872, opened 2016-04-19
Mitmproxy android

ghost:
Added functionality to push the public key element of the keypair that mitmproxy uses to Android devices during initialization of emulator and agent. This way it should be possible to use mitmproxy to dump otherwise encrypted SSL/TLS traffic.

I know it is a little redundant to specify the public key in the auxiliary.conf configuration file, however it is significantly easier than extracting just the public key from the PKCS#12 keypair on the fly. Furthermore in due time, I would like to add functionality to Cuckoo where the CA certificate for mitmproxy is generated per analysis using parameters defined in a configuration file.

Finally to help decoding of SSL/TLS in Wireshark etc, this patch contains a small fix that make mitmproxy dump TLS master secrets to the analysis directory.

### DIFF
--- a/modules/auxiliary/mitm.py
+++ b/modules/auxiliary/mitm.py
@@ -5,7 +5,6 @@
 
 import logging
 import os.path
-import shlex
 import subprocess
 import threading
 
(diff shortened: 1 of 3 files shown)


jbremer:
Sweet! This was on my todo list. I'll test it in a bit - thanks!

modules/auxiliary/mitm.py
```diff
@@ -57,22 +66,31 @@ def start(self):
...
 
         mitmerr = os.path.join(CUCKOO_ROOT, "storage", "analyses",
                                "%d" % self.task.id, "mitm.err")
 
-        self.proc = subprocess.Popen(args,
+        # Store the TLS master secrets for TLS decryption with Wireshark
+        tls_secrets_path = os.path.join(CUCKOO_ROOT, "storage", "analyses",
+                                 "%d" % self.task.id, "mitm_tls_secrets.log")
+
+        self.proc = subprocess.Popen(shlex.split(command),
```
swackhamer: arguments passed into Popen should probably be a list and not a split string in cases like this where you can control the arguments.

RuneTM:
I reverted the changes to how the mitmdump command is structured for popen, as it actually works without it, and it was probably only needed because when I started implementing this functionality I was using an old version of mitmproxy.

Also I reverted the other "time.sleep(10)" commit  from this pull request, as it is a separate pull request, sorry about that.

jbremer:
You read my mind - thanks! Will check this out later.

kuza55:
It looks like this code got added and then reverted; does anyone know what the issue was that caused it to get pulled?

jbremer:
Did anyone have the time to test this PR out by any chance? It looks quite fine, overall, but haven't had time myself.

PR #885, opened 2016-04-24
kdbg support to speed up memory dump processing

doomedraven:
processing one analysis on my machine with and without kdbg
with

real    2m24.136s
user    2m13.588s
sys     0m9.280s

without

real    2m51.553s
user    2m42.492s
sys     0m8.580s

### DIFF
--- a/conf/memory.conf
+++ b/conf/memory.conf
@@ -12,6 +12,12 @@ delete_memdump = no
 # filter: use filters to remove benign system data from the logs
 # Filters are defined in the mask section at below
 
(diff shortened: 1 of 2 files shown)


doomedraven:
@jbremer what about this one? :)

doomedraven:
@jbremer any possibility to get this merged?

jbremer:
The idea is good, but the implementation requires some improvements. Preferably we find a more generic way to perform this logic.

doomedraven:
not required in vol3, closing

PR #894, opened 2016-04-29
Update wsgi.py

eoinmiller-sfdc:
This gets rid of the missing web.settings error when you are trying to use the WSGI with Apache. Fix from older version/fork was here, still works for this:

https://github.com/spender-sandbox/cuckoo-modified/blob/master/web/web/wsgi.py

### DIFF
--- a/web/web/wsgi.py
+++ b/web/web/wsgi.py
@@ -19,6 +19,16 @@ framework.
 
 """
 import os
+import sys
+
+from os.path import join, dirname, abspath
+
+webdir = abspath(join(dirname(abspath(__file__)), '..'))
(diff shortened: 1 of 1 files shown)


jbremer:
As part of #863 this will be somewhat improved - I had already done so for `api.py`, but I'll make sure to incorporate your changes into the wsgi handler for the Web Interface.

jbremer:
This has definitely been taken care of in #863 now. Closing this PR as apparently not a lot of people use the WSGI thing at the moment and it will be available soon (tm) anyway.
In #863 the usage of the WSGI is also better documented and helped with, i.e., you can automatically generate `uwsgi` and `nginx` configuration based on your Cuckoo instance.

PR #903, opened 2016-05-05
Update community.py

copeland3300:
when running python community.py -awf, the current code leads to an exception, since os.remove expects an empty directory. 

  File "community.py", line 194, in <module>
    main()
  File "community.py", line 190, in main
    install(enabled, force, rewrite, args.archive)
  File "community.py", line 130, in install
    installdir(origin, os.path.join(CUCKOO_ROOT, folder), force, rewrite)
  File "community.py", line 80, in installdir
    os.remove(destination)
OSError: [Errno 21] Is a directory: '/home/user1/cuckoo/data/monitor/latest'

shutil.rmtree should be used to delete a directory with content in it. 

After making the change, the script completes correctly.

### DIFF
--- a/utils/community.py
+++ b/utils/community.py
@@ -77,7 +77,7 @@ def installdir(src, dst, force, rewrite, origin=[]):
             srcpath = os.path.join(src, file_name)
             if os.path.islink(srcpath):
                 if os.path.lexists(destination):
-                    os.remove(destination)
+                    shutil.rmtree(destination)
                 os.symlink(os.readlink(srcpath), destination)
                 print "Symbolic link \"%s/%s\" -> \"%s\" %s" % (
                     "/".join(origin), file_name, os.readlink(srcpath),


jbremer:
That's odd - the point of that particular `os.remove()` call is to remove the `symbolic link`, not per se to remove the entire directory that's *below* it (as in, what you find when following the symbolic link).
I'm going to have to investigate this a little bit more - thanks!

jbremer:
Thanks! Merged upon popular demand, see also #908 :-)

jbremer:
@copeland3300 @utkonos Which operating system are you guys running? The merge request that both of you requested is not breaking things on Linux-based systems.

copeland3300:
Ubuntu 16.04 server install, using python 2.7.11

jbremer:
@Und3rf10w possibly fixed the issue through #910 - could you please verify? Thanks.

copeland3300:
  File "community.py", line 199, in <module>
    main()
  File "community.py", line 195, in main
    install(enabled, force, rewrite, args.archive)
  File "community.py", line 135, in install
    installdir(origin, os.path.join(CUCKOO_ROOT, folder), force, rewrite)
  File "community.py", line 83, in installdir
    os.unlink(destination)
OSError: [Errno 21] Is a directory: '/home/user1/test/cuckoo/data/monitor/latest'

copeland3300:
I think checking 'destination' with os.path.isdir() and os.path.islink()
beforehand might be the best path forward

jbremer:
There's literally a `islink()` check before the part that throws exceptions. Also based on your last exception it would appear that the `latest` thing is a directory for you, rather than a symbolic link (assuming that the symbolic is not followed already).. but I'll have to test myself I guess by setting up the Ubuntu 16.04.

copeland3300:
Oh, I meant do os.path.islink(destination) or something like that.

But the different behavior is certainly interesting. I wonder if this has
anything to do with the way that Canonical decided to make Python 3 the
default and Python 2 a post-install option, so perhaps the libs are odd...

PR #908, opened 2016-05-09
Use shutil: os.remove is not for directories

utkonos:
os.remove raises OSError because "destination" is a directory. This error causes the rest of the community update to fail. This patch uses shutil.rmtree() which works fine. BTW: os.rmdir() will not work in this case because the directory is not empty.


jbremer:
Thanks! See also #903 which had the exact same PR :-)

PR #910, opened 2016-05-10
Symlink fixes for community.py

Und3rf10w:
Somewhat hacky solution, but in the event that ```utils/community.py``` attempts to remove a symlink using ```shutil.rmtree()```, catch the exception and use ```os.unlink()``` on it.

See #903 and #908 for reference.


jbremer:
Thanks again, @Und3rf10w!

PR #917, opened 2016-05-12
inetsim and tor integration to rooter.py

doomedraven:
### inetsim

```
wget https://googledrive.com/host/0B6fULLT_NpxMQ1Rrb1drdW42SkE/remnux-6.0-ova-public.ova
tar xvf remnux-6.0-ova-public.ova
qemu-img convert -O qcow2 REMnuxV6-disk1.vmdk remnux.qcow2 # or convert to your format
# set permament ip and configure it in conf
```

### tor
```
To install Tor, follow the guide option 2, not option 1, [here](https://www.torproject.org/docs/debian.html.en)

Once Tor is installed, edit its configuration file

$ sudo nano /etc/tor/torrc
Add the following lines to the bottom:

TransListenAddress 192.168.56.1
TransPort 9040
DNSListenAddress 192.168.56.1
DNSPort 5353
Note: 192.168.56.1 is the default address for the first VirtualBox host-only adaptor (vboxnet0). Adjust as necessary for your virtualization setup to match Cuckoo's result server IP address.

Then restart Tor:

$ sudo service tor restart
$ sudo uupdate-rc.d tor defautls
```

### DIFF
--- a/analyzer/windows/modules/packages/doc.py
+++ b/analyzer/windows/modules/packages/doc.py
@@ -17,6 +17,7 @@ class DOC(Package):
         ("ProgramFiles", "Microsoft Office", "Office14", "WINWORD.EXE"),
         ("ProgramFiles", "Microsoft Office", "Office15", "WINWORD.EXE"),
         ("ProgramFiles", "Microsoft Office 15", "root", "office15", "WINWORD.EXE"),
+        ("ProgramFiles", "Microsoft Office", "root", "Office16", "WINWORD.EXE"),
         ("ProgramFiles", "Microsoft Office", "WORDVIEW.EXE"),
     ]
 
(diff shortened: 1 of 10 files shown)


jbremer:
Thanks @doomedraven! This PR looks mostly fine to me, I'd just like to double check the `iptables` commands.
Also, I think it's time to deprecate `cuckoo.conf`'s `[routing]` and `vpn.conf` in favor of a new `routing.conf` as we're having quite some different options now and it's time to centralize those options again.
So I'll work on that somewhere this week hopefully.

doomedraven:
@jbremer you are welcome, yeap it would be great to have all routing in one file only

jbremer:
In the meantime @doomedraven, I like the dns redirection rules, but could you perhaps separate those from the `inetsim`/`tor` commands? That way they can be reused for other purposed as well (e.g., for usage with the `dnsserve.py` utility / Cuckoo App).

doomedraven:
@jbremer for sure will do it in a moment

doomedraven:
i will do new PR in a moment, with all changes needed, as don't have access to modify this last think here

PR #936, opened 2016-05-25
Allow environment variables to be used in config files.

HarryR:
This is necessary so parameters can be injected into the config file from the environment, e.g. if you're deploying with Docker, or using good devops practices in general.

```
[database]
connection = postgresql://cuckoo:%(CUCKOO_PG_PASS)s@%(CUCKOO_PG_HOST)s:5432/%(CUCKOO_PG_DB)s

# or ...

connection = %(CUCKOO_PG_DSN)s
```


jbremer:
Interesting :-) Thanks!

PR #941, opened 2016-06-02
Update guest.py

xvsfekcn:
so i think it's better to set the timeout when use requests to make a http request ,otherwise the maybe block int the (recvfrom),when use the post, i think it's also need set the timeout val.


xvsfekcn:
so how about this?

jbremer:
So at this point an initial connection has already been setup between the Host and the Guest. Therefore I don't really see why this GET request would fail. Also, does `requests` really throw `socket.error` exceptions?

jbremer:
Still sticking with my previous argument. Going to close this PR as I don't think it actually mitigates any potential issues. Please reopen if you have arguments otherwise :-)

PR #942, opened 2016-06-02
Fixes & improvements to distributed Cuckoo

HarryR:
This includes fixes and improvements to distributed cuckoo, notably:

 * Gracefully handle empty `machine=` in config
 * Re-sync instance machines when worker is re-enabled
 * Don't require separate process per cuckoo instance
 * Datetime parsing between dist & cuckoo API

These changes are required for my docker-ized Cuckoo to operate smoothly, but should be useful for everybody.

`instance.py` accepts a new parameter: `dist.handler` - this polls all enabled instances, avoiding the need to run one instance.py per worker, or updating the config when new workers are added or removed.

### DIFF
--- a/lib/cuckoo/core/rooter.py
+++ b/lib/cuckoo/core/rooter.py
@@ -14,7 +14,8 @@ from lib.cuckoo.common.config import Config
 
 cfg = Config()
 log = logging.getLogger(__name__)
-unixpath = tempfile.mktemp()
+unixpath = tempfile.mktemp(dir=cfg.cuckoo.rooter_tmp
+                               if cfg.cuckoo.rooter_tmp else None)
 lock = threading.Lock()
 
 vpns = {}
--- a/lib/cuckoo/core/rooter.py
+++ b/lib/cuckoo/core/rooter.py
@@ -40,6 +40,7 @@ def rooter(command, *args, **kwargs):
     except socket.error as e:
         log.critical("Unable to passthrough root command as we're unable to "
                      "connect to the rooter unix socket: %s.", e)
+        lock.release()
         return
 
     s.send(json.dumps({
--- a/analyzer/windows/analyzer.py
+++ b/analyzer/windows/analyzer.py
@@ -115,7 +115,7 @@ class Files(object):
 
     def dump_files(self):
         """Dump all pending files."""
-        for filepath in self.files:
+        for filepath in list(self.files):
             self.dump_file(filepath)
 
 class ProcessList(object):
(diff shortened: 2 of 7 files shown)


lib/cuckoo/core/rooter.py
```diff
@@ -14,7 +14,8 @@
 
 cfg = Config()
 log = logging.getLogger(__name__)
-unixpath = tempfile.mktemp()
+unixpath = tempfile.mktemp(dir=cfg.cuckoo.rooter_tmp
+                               if cfg.cuckoo.rooter_tmp else None)
```
HarryR: This is the only change I'm not sure about.

jbremer:
Regarding `Don't require separate process per cuckoo instance`. The goal of this was so that tasks could be submitted to several instances in parallel. These changes get rid of that feature.
I see why you'd prefer to have it all in a single process, but then we should change to something like `gevent` with one coroutine per node.

HarryR:
Changes to instance.py should be backwards compatible, so having multiple processes is still possible. However, I understand your reasoning and agree that using gevent would be a good idea. I will update the branch to use gevent, thanks :)

I also fixed a bug in `rooter()` which which cause the Cuckoo API to hang if there was an error contacting the rooter.

jbremer:
Yes, I noticed that, thanks! It is easier for us if you split those different changes among different pull requests, though. E.g., with these `gevent` changes, this PR might take a bit longer before being merged ;-)

HarryR:
Ok, when I have some spare time I'll prepare another branch with the trivial fixes so the PR can be split, namely:

 * 970f4b6
 * 19277aa
 * 9c15ea4
 * 06eff91

jbremer:
@HarryR most of the changes in this PR have been taken care of as part of #863, for which I reimplemented parts of Distributed Cuckoo. Perhaps you could take a look to see if it fits your goals? :-)
In fact, it's also using `gevent` for the centralized Distributed Cuckoo worker process.

Your https://github.com/cuckoosandbox/cuckoo/commit/970f4b64cb40b9cacf658eee7cd58a692cda6b73 got fixed as part of #982 (sorry, didn't realize you had already fixed it here), your https://github.com/cuckoosandbox/cuckoo/commit/9c15ea47a44587376952acc23672e54ae58b5ffb got fixed in 888750e31702d5b38b3a8280993cdb6a9cf0cf63 (again, didn't realize at the time you had already fixed it), and the other two fixes I'll cherry-pick soon.
I'm not sure if that leaves anything else to be merged from this PR? :)
Please let me know.

jbremer:
@HarryR Having manually fixed up all of the issues identified by you, I'm going to close this PR. Please reopen or create a new one if you find that something doesn't behave as expected. Thanks again for your submission - got some good stuff in here!

Issue #811, opened 2016-03-14
Database Error counting machines & Database error setting machine status


SparkyNZL:
Ive been collecting malware sample for a while now and have decided to run them all through cuckoobox, mainly as its really is an awesome tool. (so huge thanks to you all)

I think i might have stumbled on a bug when submitting a large number of items. >15 going to just 1 VM

The 2 Main errors i get are the following.
Database error counting machines: QueuePool limit size 5 ocerflow 10 reached, connection timed out

and

Database error setting machine status: QueuePool Limit size 5 overflow 10 reached, connection timed out, timeout 30

This is where cuckoo.py hangs  has anyone else encounter this ? solved this ? I have tried the following

Deleting and re-creating all databases
Running wiht no plugins, Re-installing it on a new system etc
both the 2.0-RC and 2.0-Dev do this

Im running nested on an ESX host HP G7 server 8T disk, 72g ram, 2 quadcore 5520 (i think) . Ubuntu 15.10  all processing is done as seperate process, mysql and MongoDB , Volitality, Snort, SSdeep, Yara, 

I would be interested to know if anyone else has the same issue our can recreate it

Cheers
Sparky

jbremer:
To be honest I don't know much about Cuckoo & MySQL, so I'd suggest using PostgreSQL instead which should do the job just fine. Until somebody else jumps in and fixes the actual error here.

SparkyNZL:
sorry about the delay in replying, I also tried postgresql and had the same error, Im currently sticking with sql lite. Cuckoo is awesome, is there a way to donate to the cause, my coding is terrible which is funny because im a RE :) perhaps i just prefer reading other peoples code :)

jbremer:
Something that *could* be related here is #946. Going to test that PR soon.
Other than that, surely there are plenty of things that you could help out with.. most importantly informing about new samples that behave unexpectedly and perhaps also analyzing them a bit to figure out why, that'd be great ;-)

SparkyNZL:
Sounds good,

This is my implementation with cuckoo

http://malware.borty.co.nz
https://moloch.borty.co.nz

l: cuckoo
p: Cuck00!

I currently have a feed from virussign.com which im feeding through the
system. so it gets a good workout, running on a HP G7 DL380 16 cores 120g
RAM and 12T of disk. Im also also running a SNORT instance on all traffic
as well

botherder:
Cool. I hope you're fine with the rest of the world logging into it now.
[...]

PR #946, opened 2016-06-05
session leak fixes

ianepanda:


jbremer:
Hi @ianepanda, are you sure about this PR? The `finally` keyword contains code that is supposed to be run *at all times*, which means both when the `try`/`catch` succeeds but also when it fails, hence this would duplicate the `session.close()` logic.

ianepanda:
Hi, @jbremer! 
imho, if `machines.filter_by(locked=False).first()` returns `None` (all filtred machnes are `locked`) then `try/except/finnaly` will not executed.

jbremer:
Oh, sorry, I incorrectly saw one indentation too much. That sounds correct, yes, thanks for the explanation and find. I'll try to do some tests and merge this PR right after :-)

SparkyNZL:
This makes sense with the issue i was having #811 great spotting :)

jbremer:
Thanks again @ianepanda :)

PR #960, opened 2016-06-13
Bugfixes

snemes:
This commit fixes 4 bugs:

- rooter.py: The iptables command allows adding the same rule multiple times, this fix modifies the disable_nat function in rooter.py to clean up all the iptables rules (ensuring that it will be actually disabled, and not only the first rule will be removed).

- analyzer.py: Due to a minor bug in the code whenever there is a problem with the auxiliary module Cuckoo will report the *previous* module's name and not the one that it was actually trying to load.

- guest.py: If the HTTP_PROXY/HTTPS_PROXY environment variables are set, the requests module will try to use them, and this is not desirable when trying to connect to a guest machine - this workaround disables the local proxy for agent connections.

- results.py: A simple formatting bug.

### DIFF
--- a/analyzer/windows/analyzer.py
+++ b/analyzer/windows/analyzer.py
@@ -600,15 +600,15 @@ class Analyzer(object):
                 aux.start()
             except (NotImplementedError, AttributeError):
                 log.warning("Auxiliary module %s was not implemented",
-                            aux.__class__.__name__)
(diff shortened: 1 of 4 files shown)


jbremer:
Sorry for only seeing this PR now - it looks good, thanks!

Issue #901, opened 2016-05-04
Memory Download doesn't work with vCenter 6 (machinery vsphere)


RedWolf74:
Hi,

in Cuckoo 2.0-RC1 the download of the memory dump (snapshot) doesn't work with vCenter 6, lower versions not testet. While downloading a file from an ESXi, the dcPath parameter (Datacenter, in which the VM resides) is optional, with vCenter it is mandatory.
After changing in vsphere.py in function _download_snapshot
param = {"dsName": datastore} to param = {"dsName": datastore, "dcPath": "DatacenterName"} 
it is working well.
Could you please add the datacenter parameter to the construction of the URL request.
This should not conflict with stand alone esxi, 'cause there is this parameter too with the default value "ha-datacenter".

Thanks.

jgajek:
Thanks for the report!  I will take a look.

jbremer:
Thanks for picking up on this issue @jgajek :)

RedWolf74:
@jgajek: thanks for taking a look at this. I know, it's a bad idea to
provide a datastore name via a config file in a vSphere env. My first
approach was to retrieve the dcPath via the pyVim api (like it is possible with i.e.
esxcli), but found no method for this in the api-docs. If you are more
familiar with the pyVim api, I suggest to get the dcPath via the api, if
it is possible.

jgajek:
Yes, it's absolutely my plan to use the API to associate the VM with the correct Datacenter.  I will send a pull request with the changes, but since I only have access to a standalone ESXi host at the moment, I will need you to test it out in a vCenter environment.

jbremer:
@RedWolf74 @certau-davidg please take a look at #978 which just got merged, thanks to @jgajek!

jgajek:
Submitted the pull request with the required changes -- can you give this a shot against a vSphere server?  I would be especially interested in whether this handles VMs within nested Datacenter folders.

RedWolf74:
I've tested against vSphere as well as stand alone ESXi, both is working well. @jgajek what do you mean with nested Datacenter folders?

jgajek:
[...]

PR #978, opened 2016-06-28
Add dcPath parameter to memory dump download URL to fix issue with vSphere deployments

jgajek:
Fix issue #901.


jbremer:
Looking sharp - thanks!

PR #988, opened 2016-06-29
KVMRemote machinery for parallel analysis using multiple hypervisors

psrok1:


botherder:
Can you elaborate on the use case?

jbremer:
What @botherder said, in which and what way could people use this? And what's the difference with regular KVM usage etc?

jbremer:
This still looks like it's going to break some setups (e.g., each machinery that doesn't take an argument for the `_connect()` function) and from what I can tell it doesn't do anything else than our current `KVM` machinery. That said I'm closing this PR. Please re-open if you think it should still be merged, but do provide additional information in that case please. Thanks again for your contribution!

PR #994, opened 2016-07-03
Analyze .jse

thejustaguy:
This adds .jse
.jse is actively used by many currently active malware families (e.g. Cerber) to download/drop the payload.


jbremer:
Sweet!

PR #999, opened 2016-07-06
Signature for identify Information Theft though a fake User-Agent

bit111:
Hi folks,
I created a simpe Signature for recognize information theft though a fake User-Agent as it happens in H-worm malware.
How can I pull the file? By attaching it in this thread?
Thanks


jbremer:
Nope :D

PR #1005, opened 2016-07-08
Fixes unable to find "application" callable in file utils/api.py

baufrecht:


jbremer:
All fixed / documented as part of #863, but I'll take it :-)

PR #1007, opened 2016-07-13
Linux guest: Add a delay when falling back to strace

drmn:
It was needed for strace to catch sample execution correctly.

### DIFF
--- a/analyzer/linux/modules/auxiliary/stap.py
+++ b/analyzer/linux/modules/auxiliary/stap.py
@@ -74,19 +74,26 @@ class STAP(Auxiliary):
         strace_start = time.time()
 
         # wait completion of strace attachment; it may take a bit of time
+        traced_fn = "strace/straced.%u" % os.getpid()
         while True:
-            try:
-                if os.path.getsize("strace/straced.%u" % os.getpid()) > 0:
-                    break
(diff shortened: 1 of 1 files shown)


jbremer:
Generally speaking I'd rather not have `time.sleep()` calls with arbitrary values. What about polling the filesize of the `strace/straced` file every second or so, and returning from the function once it's not zero anymore (i.e., `strace` has written something to it and thus is initialized)?

drmn:
Thanks for your comment. I will try to improve it reflecting your input.

jbremer:
Any luck @drmn? :-)

drmn:
Apologies the delay—2nd try!

jbremer:
Could you replace the `try`/`catch` approach with a `if not os.path.exists(...): continue`? Bit neater imo.

drmn:
Thanks for your comment. Yes I agree and fixed, and noticed/fixed that previous code would loop without sleep until file generation.

doomedraven:
this added to upcoming linux analyzer pr :)

doomedraven:
your PR included here https://github.com/cuckoosandbox/cuckoo/pull/1545 :)

drmn:
Thank you—glad to see it :)

PR #1009, opened 2016-07-14
fix overlap on wsf and html

razuz:
This fixes HTML overlap on wsf analyze as .wsf is considered as HTML file_type


jbremer:
Thanks - looking sweet!

PR #1034, opened 2016-08-10
add new basic dependencies

moros:
add dependencies:

python-lxml from pip needs: libxml2-dev libxslt1-dev
pillow needs libjpeg-dev

these -dev-packages are not in the default installation of Ubuntu 16.04


jbremer:
Thanks - although these were also already documented as part of the new `package` branch, see also eedb049bdf9de4022eb9da2ac3fd94acebb2ba8c.

jbremer:
I guess for now it'd be useful in the master branch as well, thanks!

PR #1035, opened 2016-08-11
Update api.rst

dewiestr:
When submitting an analysis through the API, its not the ID of the machine that needs to be submitted but the machine's label.


jbremer:
You're right - thanks!

PR #1039, opened 2016-08-14
very basic notification plugin for Reporting 

razuz:
In order to have external applications or scripts to collect analysis data timely it's quite costly (or even useless) to have them poking API all the time to figure out if the task has been finished or not. On the same time collecting the information from databases or other methods also doesn't work well as some analysis take more time than others and you have to keep track of their states - depending on configuration of course.

This PR basically provides a simple way to have one application/script to feed tasks from one end and this one will notifiy from other end to make something useful with the result :) It passes down `results.info` in json as `data` and the host `identifier` so that notification receiver would know to which host from the cluster reported it. Having the module to POST more data doesn't make much sense as all of it can be grabbed via other methods when suitable. However it can be made configurable in case it makes sense.

@jbremer not sure but maybe it makes sense to abstract `def default(obj):` as it's used here the same way as in jsondump (https://github.com/cuckoosandbox/cuckoo/blob/master/modules/reporting/jsondump.py#L15-L20)

### DIFF
--- a/conf/reporting.conf
+++ b/conf/reporting.conf
@@ -56,3 +56,16 @@ enabled = no
 # moloch_capture = /data/moloch/bin/moloch-capture
 # conf = /data/moloch/etc/config.ini
 # instance = cuckoo
+
+[notification]
+# Notification module to inform external systems that analysis is finished.
+# You should consider keeping this as very last reporting module.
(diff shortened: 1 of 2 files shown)


jbremer:
Thanks again!

PR #1046, opened 2016-08-17
Integrate DroidBot into Cuckoo Sandbox

tinduong94:
This is a pull request to complete Google Summer of Code 2016 - Integrating DroidBot into Cuckoo Sandbox.


jbremer:
Just to be clear, it is necessary for `DroidBot` to run *in* the emulator, yes? We can't run it from the outside through `adb` or so?

analyzer/android/analyzer.py
```diff
@@ -38,41 +38,45 @@ def get_options(self):
...
 
         # Parse the analysis configuration file generated by the agent.
-        self.config = Config(cfg="analysis.conf")
+        self.config = Config(cfg="analysis.json")
 
         # We update the target according to its category. If it's a file, then
         # we store the path.
         if self.config.category == "file":
-            self.target = os.path.join("/data/local/tmp", str(self.config.file_name))
+            self.target = os.path.join("/tmp", str(self.config.file_name))
```
jbremer: Does Android have a `/tmp`?

agent/agent.py
```diff
@@ -122,12 +121,20 @@ def add_config(self, options):
...
+                    config["analysis"][key] = value
+                else:
+                    try:
+                        value = value.__str__()
+                    except AttributeError:
+                        value = ""
+                    config["analysis"][key] = value
 
-            config_path = os.path.join(self.analyzer_folder, "analysis.conf")
+            config_path = os.path.join(self.analyzer_folder, "analysis.json")
```
jbremer: We can't expect users to update their Cuckoo Agent, as that hasn't been updated in years. As such a fallback to `config.conf` will be required if `config.json` couldn't be found (by the Cuckoo Analyzer). This will most likely require writing both `config.json` and `config.conf` files with the Analyzer attempting `config.json` first, etc.

analyzer/android/lib/api/droidbot/common.py
```diff
@@ -0,0 +1,46 @@
+# Copyright (C) 2012-2015  Diego Torres Milano
```
jbremer: Just to be clear, it is necessary for `DroidBot` to run *in* the emulator, yes? We can't run it from the outside through `adb` or so? And then, would it be possible to install a `Droidbot` Python package during the device/emulator setup, rather than embedding it directly in Cuckoo?

lib/cuckoo/common/config.py
```diff
@@ -65,15 +65,28 @@ def get(self, section):
 
 def parse_options(options):
     """Parse the analysis options field to a dictionary."""
-    ret = {}
-    for field in options.split(","):
-        if "=" not in field:
-            continue
+    try:
```
jbremer: Could you write some unit tests to verify that this new logic is both backwards compatible as well as working with the new purpose in mind?

lib/cuckoo/core/guest.py
```diff
@@ -343,21 +343,23 @@ def upload_analyzer(self, monitor):
...
+                except AttributeError:
+                    value = ""
+                config["analysis"][key] = value
 
         data = {
-            "filepath": os.path.join(self.analyzer_path, "analysis.conf"),
+            "filepath": os.path.join(self.analyzer_path, "analysis.json"),
         }
-        self.post("/store", files={"file": "\n".join(config)}, data=data)
+        self.post("/store", files={"file": json.dumps(config)}, data=data)
```
jbremer: You're communicating with the new Cuckoo Agent here, but made changes for the Old Agent above? Has this functionality been tested?

PR #1062, opened 2016-08-30
add suricata check to Signatures class

doomedraven:
This is useful to check for family names in alerts

### DIFF
--- a/lib/cuckoo/common/abstracts.py
+++ b/lib/cuckoo/common/abstracts.py
@@ -1047,7 +1047,7 @@ class Signature(object):
         @return: True/False
         """
         res = False
-        for alert in self.results.get("suricata", {}).get("alerts", []):
+        for alert in self._caller.results.get("suricata", {}).get("alerts", []):
             if re.findall(pattern, alert.get("signature", ""), re.I):
                 res = True
                 break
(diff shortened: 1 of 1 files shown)


jbremer:
Hmm, I must have missed this one earlier. Should we expand a little bit on it? E.g., being able to also search for the SID, etc? Or supporting exact matches rather than regexes?

doomedraven:
we could add support for SIDs, well we could use exact search in alert msg, but it wont be so advanced, regex gives more flexibility, and this is very useful pattern to detect malware family if there no signature for that family or just double verifications, and to extract data in signature/plugins it helps a lots as one of the steps to detect family X

lib/cuckoo/common/abstracts.py
```diff
@@ -1041,6 +1041,18 @@ def check_url(self, pattern, regex=False, all=False):
...
 
+    def check_suricata_alerts(self, pattern):
+        """Check for pattern in Suricata alert signature
+        @param pattern: string or expression to check for.
+        @return: True/False
+        """
+        res = False
+        for alert in self._caller.results.get("suricata", {}).get("alerts", []):
+            if re.findall(pattern, alert.get("signature", ""), re.I):
+                res = True
```
jbremer: Please replace this logic by a `return True` here and a `return False` at the end of the function and then I'll merge it!

doomedraven:
new PR here as i can't edit this one https://github.com/cuckoosandbox/cuckoo/pull/1262

Issue #1067, opened 2016-09-03
Windows Analyzer: Time drift bug


ChriDr:
Hi,

I think there is a Bug within the analyzer/windows/lib/core/startup.py file . When an analysis starts, the system time will be set to the time when the task was submited. The timzone information is also included in that time, it is not just an utc time.

Therfore when the set_clock(clock) method is used, the time will be interpreted as utc time without timezone information. Because KERNEL32.SetSystemTime(ctypes.byref(st)) set the system as utc time. Then the time jumbs e.g 2 hours ahead.

Instead of SetSystemTime, the SetLocalTime function should be used.

Best regards,

jbremer:
Thanks for that insight. Would you like to do a small pull request for that? :-)

ChriDr:
Sure :).  I will do it today or tomorrow

jbremer:
Resolved as per your PR, thanks again.

PR #1068, opened 2016-09-03
Using SetLocalTime instead of SetSystemTime

ChriDr:
As described in [https://github.com/cuckoosandbox/cuckoo/issues/1067](url), the time is misinterpreted by windows when SetSystemTime is  being used.


jbremer:
Thanks!

ChriDr:
Your are welcome!

PR #1070, opened 2016-09-05
Package

jpalanco:
I changed os.getlogin() by pwd.getpwuid(os.getuid())[0]

The reason is because os.getlogin() doesn't work without tty. For example, cuckoo api will not work under docker.

The os.getlogin() documentation recommends the use of pwd.getpwuid(os.getuid())[0] 

References:https://docs.python.org/2/library/os.html#os.getlogin


jbremer:
Thanks.

jbremer:
@jpalanco Perhaps related to your interests, https://github.com/cuckoosandbox/cuckoo/commit/0594c2ccb71e8788c54a4106235e827a689e3f53. Thanks again!

PR #1072, opened 2016-09-06
office publisher package :)

doomedraven:
just tested, works perfectly, uniq think which i dislike is to have
list of values per reg key, probably we could do something like
versions = [12.0, 15.0,16.0, etc ] and loop them so something similar


jbremer:
Thanks! I've manually merged it to avoid the many `merge remote-tracking branch` commits https://github.com/cuckoosandbox/cuckoo/commit/7555b2a995446ff166e04c8ae7cf429e25f9c82d.

PR #1087, opened 2016-09-12
api: submitt download-file-from-url

tcert:
added Route /tasks/create/downloadfile
requires POST parameter 'url'

Submitt  a file to the queue, the file will be downloaded from the url (for example:  Owncloud, FileShare, FileRepo, etc...)

### DIFF
--- a/modules/machinery/vsphere.py
+++ b/modules/machinery/vsphere.py
@@ -323,16 +323,27 @@ class vSphere(Machinery):
                 (name, vm.summary.config.name)
             )
 
-        datakey = filespec = None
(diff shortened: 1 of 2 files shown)


jbremer:
@tcert Thanks for the suggestion. We've considered implementing this in the past, however, we foresee security implications with downloading resources off of internal hosts on the network.
If there is a way to mitigate this using the `requests` library, I'll definitely consider implementing this feature (provided that appropriate unit tests are added as well).

jbremer:
Closing for reasons mentioned above.

PR #1089, opened 2016-09-15
I made the following changes to be able to successfully analyse andro…

p2rate:
…id apps:

I have made changes to avd.py so that it does not create a copy of original AVD.
it runs the same AVD provided by refrence_machine var. this will allow us to
run the AVD from snapshot. I also disabled the headless mode because in my
experince AVD cannot be run from snapshot in headless mode.
the reason to run AVD from snapshot is that if you root an AVD and restart it,
superuser app failes to give superuser access to app(although the su binary
is in place and given the excutable permission). to fix this problem you have
to do the rooting process all over.but since running AVD from snapshot does
not go through boot process, we won't have that problem.

I have also made changes to lib/cuckoo/common/objects.py. I tried to analyse
some apk files. get_apk_entry was not able to find the package name and the
name of the MainActivity of most of the apps I tried to analyse.
I installed apktool and made get_apk_entry use this tool. apktool extracts
an apk file and I read the AndroidManifest.xml file. AndroidManifest.xml
contains the name of the package.this works for all apk files.

I also made changes to adb.py. I changed execute_sample function. this way
you don't need to know the name of the MainActivity.

other changes was made as a result of the changes I just described.

### DIFF
--- a/analyzer/android/lib/api/adb.py
+++ b/analyzer/android/lib/api/adb.py
@@ -23,14 +23,19 @@ def install_sample(path):
 
     log.info("Installed sample: %r", output)
 
-def execute_sample(package, activity):
+def execute_sample(package):
     """Execute the sample on the emulator via adb"""
     try:
-        package_activity = "%s/%s" % (package, activity)
-        args = [
-            "/system/bin/sh", "/system/bin/am", "start",
-            "-n", package_activity,
+                args = [
+	    "/system/bin/sh",
+            "/system/bin/monkey",
+	    "-p", 
+	    "%s" % package,
+	    "-c",
+            "android.intent.category.LAUNCHER",
+	    "1",
         ]
+
         output = subprocess.check_output(args)
     except subprocess.CalledProcessError as e:
         log.error("Error executing package activity: %r", e)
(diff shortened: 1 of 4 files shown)


jbremer:
There are workarounds for the root after reboot issue (documented somewhere in the `cuckoo-droid` Github page - please feel free to PR Android documentation). Copying the reference AVD is required in case one wants to run multiple AVDs in parallel. Furthermore, I'm sure there's a better way to read in `AndroidManifest.xml` than running two external commands :-)
That said, as-is we can't merge this PR.

p2rate:
to ROOT an AVD, we must put the SU file in /system and give it executable permission. but changes do not withstand a restart because the /system filesystem is mounted from a file in android sdk.
the solution in cuckoo-droid documentation does address the problem I just described but not the following:
even if you successfully root an AVD and make the changed to /system permanent, superuser app does not work properly. this is despite the existence of SU file, which has excutable permission.
**you may test it yourself and learn the truth the hard way.** :D 

as for copying the AVD: as I'm aware cuckoo does not support multiple AVDs at the time.

 at last: I know running two external commands may not be as clean as it could be, but at least it works.

PR #1092, opened 2016-09-16
Add restart functionality for physical emulation

marklech:
When conducting emulation with the physical machinery, the physical guest machine 
has to be rebooted prior to subsequent analyses. This is currently done by issuing an 
RPC to the Windows RPC server, which is sent via the SMB protocol. At times, there are 
some underlying issues in SMB that prevent the RPC to be executed in the intended manner, 
and as a result the machine doesn't reboot, leaving it in an unclean state for further analysis. 
This can be over come by adding the option of rebooting via Cuckoo agent's XML-RPC server
by registering the aforementioned functionality as a procedure on the server. An additional 
change is also made to physical.py to accomplish this.

### DIFF
--- a/agent/agent.py
+++ b/agent/agent.py
@@ -174,6 +174,15 @@ class Agent(object):
         self.current_status = STATUS_RUNNING
         return self.analyzer_pid
 
+    def restart(self):
+        """ Restart the machine (for physical machine analysis). 
+        """
+        try:
+            subprocess.call(["shutdown","-r"])
+            return True
+        except: 
+            return False
+        
     def complete(self, success=True, error="", results=""):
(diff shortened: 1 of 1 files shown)


jbremer:
Our old Agent hasn't been updated in years (backwards compatibility) and therefore we can't really expect users to update it. Also there's no way of knowing if the `restart` function is present.
Therefore I'd rather you use the new Agent in which this functionality is already available (through the `/execute` API endpoint) https://github.com/jbremer/agent/blob/master/agent.py. In an upcoming release we'll be updating the default Agent to this new Agent, while still keeping backwards compatibility of course (as it is right now).

Issue #948, opened 2016-06-07
Whitelisting URL`s not working


donatasiv:
Hi folks,

I wanted to whitelist some url in report, because a lot of url entries generates MS Office (when inspecting .doc files) which are actually not relevant or malicious. What I did:
1. updated cuckoo_root/data/whitelist/domain.txt file whit urls`s I dont want to see in report.
2. updated processing.conf
    [network]
    enabled = yes
    dnswhitelist = yes

But nothing happens. Anything I`m missing?

jbremer:
This is currently not yet a feature, hopefully to be fully implemented *soon*'ish. If you'd be up for creating a pull request on this topic, that'd be great of course.

razuz:
so that should be resolved by #1094

jbremer:
Closing as resolved for reasons described above by @razuz.

PR #1094, opened 2016-09-17
basic implementation of domain whitelisting

razuz:
Here is basic implementation of domain whitelisting which also solves issue #948 

not sure but do you think it should be configurable in order to enable/disable it ?

### DIFF
--- a/conf/processing.conf
+++ b/conf/processing.conf
@@ -58,6 +58,11 @@ maxioc = 100
 [network]
 enabled = yes
 
+# Allow DNS responses from your configured DNS server for whitelisting to
+# deactivate when responses come from some other DNS
+# Can be also multiple like : 10.0.0.1,10.0.0.2
+allowed-dns = 
+
 [procmemory]
 # Enables the creation of process memory dumps for each analyzed process right
 # before they terminate themselves or right before the analysis finishes.
--- a/conf/processing.conf
+++ b/conf/processing.conf
@@ -58,6 +58,9 @@ maxioc = 100
 [network]
 enabled = yes
 
+# Allow domain whitelisting
+whitelist = no
+
 # Allow DNS responses from your configured DNS server for whitelisting to
 # deactivate when responses come from some other DNS
 # Can be also multiple like : 10.0.0.1,10.0.0.2
--- a/conf/processing.conf
+++ b/conf/processing.conf
@@ -59,7 +59,7 @@ maxioc = 100
 enabled = yes
 
 # Allow domain whitelisting
-whitelist = no
+whitelist-dns = no
 
 # Allow DNS responses from your configured DNS server for whitelisting to
 # deactivate when responses come from some other DNS
--- a/conf/processing.conf
+++ b/conf/processing.conf
@@ -63,7 +63,7 @@ whitelist = no
 
 # Allow DNS responses from your configured DNS server for whitelisting to
 # deactivate when responses come from some other DNS
-# Can be also multiple like : 10.0.0.1,10.0.0.2
+# Can be also multiple like : 8.8.8.8,8.8.4.4
 allowed-dns = 
 
 [procmemory]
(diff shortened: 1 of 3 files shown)


modules/processing/network.py
```diff
@@ -111,6 +111,21 @@ def __init__(self, filepath):
...
         self.results = {}
+        # List containing all whitelist entries.
+        self.whitelist = self._build_whitelist()
+
+    def _build_whitelist(self):
+        result = []
+        whitelist_path = os.path.join(
+            CUCKOO_ROOT, "data", "whitelist", "domain.txt"
+        )
+        f = open(whitelist_path, 'rb')
```
doomedraven: i think it would be more better/pretty
 change this
```
+        f = open(whitelist_path, 'rb')
 +        while True:
 +            line = f.readline()
 +            if not line: break
 +            result.append(line.split('\n')[0])
 +        
```

to this

```
f = open(whitelist_path, 'rb').readlines()
f = [whitelisted.strip() for whitelisted in f]
```

razuz: @doomedraven sure - I'll verify that also just in case later today and commit when load drops off my box

jbremer:
I still think that each domain name should consist a set of known IP addresses that are whitelisted, and that we can only whitelist an entry if it was requested to a known legitimate DNS server (e.g., `8.8.8.8`).
Might require some additional work, as well as unit tests, but up until then ;-)

razuz:
alright ... so as discussed : 
* need a place in conf where to define "known good ns server" for current environment
* in case ns response comes from other ns server than good one - the record should not be whitelisted

I'll figure that out

modules/processing/network.py
KillerInstinct: why not just use with open() ?

razuz:
@jbremer I guess you meant something like that :) 
Check it out :)

conf/processing.conf
```diff
@@ -58,6 +58,14 @@ maxioc = 100
 [network]
 enabled = yes
 
+# Allow domain whitelisting
+whitelist = no
+
+# Allow DNS responses from your configured DNS server for whitelisting to
+# deactivate when responses come from some other DNS
+# Can be also multiple like : 10.0.0.1,10.0.0.2
+allowed-dns =
```
jbremer: I think a default of `8.8.8.8,8.8.4.4` would be useful?

modules/processing/network.py
```diff
@@ -111,6 +113,39 @@ def __init__(self, filepath):
...
+            log.debug("Whitelisting Disabled.")
+            return False
+
+        return True
+
+    def _build_known_dns(self):
+        """Build known DNS list."""
+        _known_dns = self.options.get("allowed-dns")
+        if _known_dns is not None:
+            return _known_dns.split(",")
```
jbremer: Needs some more stripping in case the user specifies `8.8.8.8, 8.8.4.4` instead of without the whitespace.

modules/processing/network.py
```diff
@@ -371,6 +410,14 @@ def _add_dns(self, udpdata):
                 # TODO: add srv handling
                 query["answers"].append(ans)
 
+            if ( self.whitelist_enabled and 
+                (conn.get("src") in self.known_dns or conn.get("dst") in self.known_dns) and  
+                q_name in self.whitelist 
+            ):
```
jbremer: Could you reformat this a little bit? Perhaps cleaner to assign temporary `src` and `dst` variables outside of the if statement.

modules/processing/network.py
jbremer: Also, right now this will probably return a list with one empty item for the current default configuration, i.e., `[""]`.

modules/processing/network.py
razuz: refactored the check and it's more dynamic now as moved to a function. Also adds option that if list is empty then it means accept all

modules/processing/network.py
razuz: that's fixed

conf/processing.conf
razuz: I don't think it should have default value as empty value == all

PR #1098, opened 2016-09-19
a bit more improved publisher package

doomedraven:
https://support.office.com/en-us/article/Command-line-switches-for-Publisher-2007-f9ba86d1-7d44-45a9-80ac-41cb158361df

/o filename

Starts Publisher and opens an existing file.
Example    To open a file, type the following at the command prompt:

/o filename

NOTE: The filename variable can identify the file by the path to the file, by a URL, or by the file path to the Windows Explorer shortcut (*.LNK) to the file or URL. If the file name has spaces in it, enclose the complete name in quotation marks. For example, /t "Contoso Business Card.pub"


jbremer:
Thanks, you're correct!

PR #1099, opened 2016-09-20
Packagefixes for ie, doc, exe

razuz:
This PR creates slightly more comfortable environment in Windows 7 for samples. One of main things is Word 2010 (office 14) fixes for registry which I was doing by hand before and enables droppers to bring better results. Excel etc coming up later.

As discussed before - in some point lets move them into profiles or something like that in order to be more dynamic :)

### DIFF
--- a/analyzer/windows/modules/packages/exe.py
+++ b/analyzer/windows/modules/packages/exe.py
@@ -22,7 +22,7 @@ class Exe(Package):
                 # "Set Up Windows Internet Explorer 8"
                 "DisableFirstRunCustomize": 1,
 
-                ### added
+                # Make Windows environment more loose and clean
                 "NoProtectedModeBanner": 1,
                 "NoUpdateCheck": 1,
                 "Start Page": "about:blank",
--- a/analyzer/windows/modules/packages/exe.py
+++ b/analyzer/windows/modules/packages/exe.py
@@ -11,6 +11,28 @@ from lib.common.abstracts import Package
 class Exe(Package):
     """EXE analysis package."""
 
+    REGKEYS = [
+        [
+            HKEY_LOCAL_MACHINE,
+            "Software\\Microsoft\\Security Center",
+            {
+                # "Would you like Internet Explorer as default browser?"
+                "Check_Associations": "no",
+
+                # "Set Up Windows Internet Explorer 8"
+                "DisableFirstRunCustomize": 1,
+
+                ### added
+                "NoProtectedModeBanner": 1,
+                "NoUpdateCheck": 1,
+                "Start Page": "about:blank",
+                "Enable Browser Extensions": "yes",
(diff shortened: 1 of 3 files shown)


analyzer/windows/modules/packages/exe.py
```diff
@@ -11,6 +11,28 @@
 class Exe(Package):
     """EXE analysis package."""
 
+    REGKEYS = [
+        [
+            HKEY_LOCAL_MACHINE,
+            "Software\\Microsoft\\Security Center",
+            {
+                # "Would you like Internet Explorer as default browser?"
```
doomedraven: can you explain what IE keys affect .exes?

analyzer/windows/modules/packages/ie.py
```diff
@@ -92,6 +134,14 @@ class IE(Package):
                 # "Internet Explorer - Security Warning"
                 # "The publisher could not be verified."
                 "CheckExeSignatures": "no",
+                "RunInvalidSignatures": 1
+            },
+        ],
+        [
+            HKEY_CURRENT_USER,
+            "Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings\\Zones\\0",
```
doomedraven: this will disable identifier zones ?

razuz: @doomedraven no ... this is realted to ActiveX ref: https://support.microsoft.com/en-us/kb/182569

analyzer/windows/modules/packages/exe.py
razuz: the point here with IE stuff is to lower the noise when you run (for example) ransomware binaries (not droppers) which bring up IE in the end or during analysis. As mentioned in the PR - in long term it makes much more sense to flip code around and build profiles for vm-s rather than inject registry changes in packages.

doomedraven: yah is what i did, preconfigured vm correctly :) but good work :)

razuz: that's true, but I'd like to have this bit automated aswell (I guess you might also) :) It makes more sense and buys me 30 min if I need to rebuild whole sandbox from scratch ;)

doomedraven: use templates vms ;)

razuz: that's one option however then you have to manage your template(s) also ;) anyways it doesn't really matter - it's mostly about about the taste how people like to manage their infra. As vmcloak is also quite dynamic then my wild guess is that "profiles" could be the optimal sweetspot as they can be reused on both.

jbremer:
I believe this PR is ready for merge?

razuz:
not really ... tweaking parameters as I go still

PR #1100, opened 2016-09-20
Fix agent.py timeout error caused by tcpdump process.

consen:
When I ran cuckoo, encountered many time out error:

```
2016-09-20 14:04:27,790 [lib.cuckoo.core.guest] DEBUG: vm-117: error retrieving status: timed out
2016-09-20 14:04:28,799 [lib.cuckoo.core.scheduler] ERROR: The analysis hit the critical timeout, terminating.
```

```
2016-09-20 14:08:58,540 [lib.cuckoo.core.scheduler] ERROR: vm-111: the guest initialization hit the critical timeout, analysis aborted.
```
Finally I found it's caused by tcpdump process, which inherits parent process's opened connections:

```
tcp      259      0 192.168.56.1:44116            192.168.56.106:8000           CLOSE_WAIT  24558/tcpdump       
tcp        0      0 192.168.56.1:50038            192.168.56.113:8000           ESTABLISHED 24356/tcpdump
```

Set `closed_fds` to`Ture`, cuckoo works well.

### DIFF
--- a/modules/auxiliary/sniffer.py
+++ b/modules/auxiliary/sniffer.py
@@ -88,7 +88,7 @@ class Sniffer(Auxiliary):
 
         try:
             self.proc = subprocess.Popen(
-                pargs, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+                pargs, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True
             )
         except (OSError, ValueError):
             log.exception(


jbremer:
Interesting! I'm wondering what would be the best way to unittest this. Also note that `close_fds` does not exist on Windows-based platforms (for the host), so we'll be needing special handling there (same goes for `close_fds` usage in the VirtualBox machinery module).

consen:
Yes, Windows needs to be handled specially.

>If close_fds is true, all file descriptors except 0, 1 and 2 will be closed before the child process is executed. (Unix only). Or, on Windows, if close_fds is true then no handles will be inherited by the child process. Note that on Windows, you cannot set close_fds to true and also redirect the standard handles by setting stdin, stdout or stderr.

I have tested this on Linux, no `time out` error.

But I have not known the real reason, I just found when a new virtual machine restored, the tcpdump inherited ESTABLISHED connections may cause agent.py hang and no response for a long time, until the tcpdump exits.

jbremer:
Sounds good to me. Will create a separate issue for unit testing & Windows support. Thanks again!

consen:
I have found the real reason, and explains the details about this bug on my blog :) https://consen.github.io/2016/09/20/a-network-timeout-bug/

When the child process inherits the parent process's socket connection, even though parent process close the socket, the socket is still opened in the child process, so the SimpleXMLRPCServer(single-threaded) will be hanging and no response.

PR #1102, opened 2016-09-21
Use named temp file for create/file in API

RicoVZ:
This PR changes the file util method used to store a analysis binary if it is submitted through the API.
The recent change to Files.temp_put introduced the unwanted behavior of an analysis binary being created at wherever a Cuckoo instance was started from, instead of it being created in cuckoo_tmp.

Changing the used method to store the analysis binaries resolves this issue.


jbremer:
Sweet, thanks!

jbremer:
This fix is now partially unittested by https://github.com/cuckoosandbox/cuckoo/commit/afd4b9a8a2c4b2594aff65ee1f8a553da3618a31 :-)

PR #1103, opened 2016-09-21
Merging CuckooML into Cuckoo

So-Cool:
Merging work done on CuckooML project during Google Summer of Code 2016 into main Cuckoo development repository.


cuckoo.py
```diff
@@ -24,6 +24,7 @@
     from lib.cuckoo.core.startup import init_logging, init_modules
     from lib.cuckoo.core.startup import init_tasks, init_yara, init_binaries
     from lib.cuckoo.core.startup import init_rooter, init_routing
+    from modules.cuckooml.cuckooml import init_cuckooml
```
jbremer: We should move `cuckooml.py` to the `modules/processing/` directory altogether, as it is processing information on a per analysis basis.

lib/cuckoo/common/virustotal.py
```diff
@@ -168,40 +398,227 @@ def file_scan(self, filepath):
         r = self._request_json(self.FILE_SCAN, data=data, files=files)
         return dict(summary=dict(permalink=r.get("permalink")))
 
+    def detect_platform(self, tokens):
```
jbremer: Rather than using AV signatures, we may consider doing this through simple static analysis of a given file. In that case, it probably makes most sense to move this logic to the following library, where it is easier to unit tests (being a small library and all that): https://github.com/jbremer/sflock.

modules/cuckooml/cuckooml.py
```diff
@@ -0,0 +1,1677 @@
...
+import json
+import os
+import re
+import sys
+import time
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import seaborn as sns
+from hdbscan import HDBSCAN
```
jbremer: These imported library will have to be added to `requirements.txt` so that users will install them and may actually use this functionality. Also, `try` / `except ImportError` statements should be placed here in case these libraries have not been installed (due to their nature, this is quite likely).

In the future, through the Python `package` (see also #863), we may be able to force these packages on to users, if not by explicitly enabling the `machine learning` capabilities.
That said, are there any special OS package requirements when installing these Python libraries? E.g., some libraries that compile native Python extensions require the `libpython-dev` package in Ubuntu.

modules/cuckooml/cuckooml.py
```diff
@@ -0,0 +1,1677 @@
...
+        else:
+            ml.save_clustering_results(loader)
+
+
+class ML(object):
+    """Feature formatting and machine learning for Cuckoo analysed binaries.
+    All functions marked with asterisk (*) were inspired by code distributed
+    with "Back to the Future: Malware Detection with Temporally Consistent
+    Labels" by Brad Miller at al."""
+    SIMPLE_CATEGORIES = {
```
jbremer: Good stuff! Do you think it makes sense to move this *knowledge* (i.e., the `has_authenticode` Signature belongs to the `properties` category) to the Signatures themselves (https://github.com/cuckoosandbox/community)?

modules/cuckooml/cuckooml.py
```diff
@@ -0,0 +1,1677 @@
...
+                my_features[i][":sign:" + j] = 1
+            for j in self.__handle_string(features[i]["LegalCopyright"]):
+                my_features[i][":sign:" + j] = 1
+            for j in self.__handle_string(features[i]["InternalName"]):
+                my_features[i][":sign:" + j] = 1
+            for j in self.__handle_string(features[i]["CompanyName"]):
+                my_features[i][":sign:" + j] = 1
+
+            # Extract packer
+            patterns = [r"Armadillo", r"PECompact", r"ASPack", r"ASProtect",
```
jbremer: If you could move this to a class variable or somewhere else outside the actual function, that'd improve the code a bit :)

modules/cuckooml/cuckooml.py
```diff
@@ -0,0 +1,1677 @@
...
+                my_features[i][":file:touch:" + f] = 1
+            for f in features[i]["file_deleted"]:
+                my_features[i][":file:touch:" + f] = 1
+            for f in features[i]["file_copied"]:
+                my_features[i][":file:touch:" + f] = 1
+            for f in features[i]["file_renamed"]:
+                my_features[i][":file:touch:" + f] = 1
+            # TODO: better binning (linear not logarithmic)
+            # File numbers
+            my_features[i][":file:count:all"] = \
```
jbremer: Surely we can turn this into a loop of sorts?

modules/cuckooml/cuckooml.py
```diff
@@ -0,0 +1,1677 @@
...
+        plt.title(fig_name + " (lr:" + str(learning_rate) + ")")
+        if self.context == "notebook":
+            plt.show()
+        else:
+            plt.savefig(fig_name + "_" + str(learning_rate) + ".png",
+                        bbox_inches='tight', pad_inches=1.)
+            plt.close()
+
+
+    def save_dataset(self, filename="custom_dataset.csv", features=None, \
```
jbremer: Is this data currently exported anywhere? E.g., exporting it to the storage directory of the current analysis would make sense. You'll be able to do this quite easily when moving this to a `Processing` module.

modules/cuckooml/cuckooml.py
```diff
@@ -0,0 +1,1677 @@
...
+            self.clustering["dbscan"] = {
+            # self.clustering["dbscan"][clustering_hash] = {
+                "eps":eps,
+                "min_samples":min_samples,
+                "clustering":pd.DataFrame(dbscan.labels_, index=features.index,
+                                        columns=["label"])
+            }
+
+
+    def cluster_hdbscan(self, features=None, min_samples=1, \
```
jbremer: Is this a resource expensive call?

hgascon: @So-Cool where is actually the module hdbscan?

So-Cool: @hgascon It's a separate package that you can get from pip.

So-Cool: @jbremer it largely depends on the number of datapoints to be clustered; but in general (with, say, reasonable number of datapoints) it shouldn't be resource expensive.

lib/cuckoo/common/virustotal.py
So-Cool: @jbremer How do you want me to go about it?

modules/cuckooml/cuckooml.py
So-Cool: @jbremer As far as I can understand you want to detach the *inspired* code from cuckooml and put it in the `cuckoosandbox/community` repository; is that correct?
If so it would probably only make it more difficult to use, wouldn't it?

modules/cuckooml/cuckooml.py
So-Cool: @jbremer it is saved as CSV file into current directory. How can I save it by default to the current analysis directory?

modules/cuckooml/cuckooml.py
So-Cool: @jbremer see 	2bf7d28

modules/cuckooml/cuckooml.py
So-Cool: @jbremer see 	0381d14

cuckoo.py
So-Cool: @jbremer see 	04aa629

modules/cuckooml/cuckooml.py
So-Cool: @jbremer see 	2e5c87e

So-Cool:
@jbremer @hgascon I tried to address as many pointed out issues as possible. What do you think?

So-Cool:
@jbremer I've fixed couple more things; the build has succeeded. Could you provide me with some pointers on further cuckooml improvements?

I've noticed that issues referenced in my commits point to `cuckoosandbox/cuckoo` repository rather than `honeynet/cuckooml`. Should I edit all these commit messages to explicitly point to `honeynet/cuckooml` issues?

prats84:
Is this merged into a branch ?

So-Cool:
I believe it's not merged yet; there are still couple of changes that need to be made.

jbremer:
@So-Cool I'd like to merge all of your work in 1-2 months from now (right after our upcoming release). Could you maybe give me an up-to-date overview on what all of your changes do, what still has to be done, and also everything we'd need to know & document? Thanks :)

So-Cool:
I need to look into it again to give you more precise answer but in general the changes to the native cuckoo files are restricted to updating `requirements.txt` and adding cuckooml command line argument in `cuckoo.py`.  
The rest of the cuckooml is fairly independent. Is needs some small tweaks (see issues in the cuckooml repo) and testing (especially after I have moved it to `modules/processing`), but other than that it should be fine.  
All you need to know and document will probably be formalised when I start testing it in preparation for merge. At the moment I'm thinking that going through GSoC'16 blog posts will give you the best insight.

jbremer:
Guess I won't be merging this, provided I don't know who would be maintaining it and (unfortunately) don't see a direct use-case.

PR #1106, opened 2016-09-23
Vt intelligence download

doomedraven:
many users have only intelligence key, not private, this will allows them download :)

### DIFF
--- a/cuckoo/core/startup.py
+++ b/cuckoo/core/startup.py
@@ -546,7 +546,7 @@ def init_vt_download():
                     # no permission in private api
                     if r.status_code == 403:
                         exception = True
-                    # no permission in inteligence api
+                    # no permission in intelligence api
                     # if user have perissions to download from vt intelligence it will return file
                     elif r.status_code == 200 and r.content[:20] == "\n\n\n\n<!DOCTYPE html P":
                         exception = True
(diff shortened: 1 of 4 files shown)


jbremer:
Sweet! Needs a bit more work though.
* Verify the `apikey` `type` somewhere during startup.
* Show an error if somebody attempts to download with a `public` VT key.
* Resolve syntax error (no colon after `else`).

jbremer:
Regarding the error reporting, I'm having my guy look at that. Still has to be done though :-)

doomedraven:
cool :)

so leave this open or close?

jbremer:
If you could do the `apikey` `type` verification at startup (somewhere in `startup.py`) then I'll merge it :)

doomedraven:
ok, looking where it would be correct to put the check :)

doomedraven:
hm thinking a bit, how you want check that in startup?
we can check if `key` is different and if `type` is `private/intelligence`
but there no option like enable downloads

what do you think about this
* i'm adding option to config `download = no`
* if that option is set to `yes`
* it will do req to vt will say to try download pafish
* in response messege we will have the status code and msg which will say if we can download or not
* if we can't download, raise exception with message about incorrect key/type

doomedraven:
@jbremer check this :)

jbremer:
I like the idea but it needs some work. I don't think we should do the checking for every Cuckoo startup. This PR also needs some unit tests.

doomedraven:
i can remove check on startup easily, what you would change/improve? i will do a unit test next week

PR #1108, opened 2016-09-23
Fix registry.py

ianepanda:
RegSetValueExW needs length of value in _bytes_, not unicode characters


jbremer:
The trailing nullbyte should be set as well, according to the documentation. Could you please update your PR to do that as well?

jbremer:
Hm, actually, I had already performed a similar fix as part of the WIP on kernel driver integration, and in the meanwhile this has been merged into the latest version of the `package` branch. See also https://github.com/cuckoosandbox/cuckoo/commit/d13bdeda7e742bdd466e7f14922ccb2d63ea40df#diff-a21bb9392789944be66ad2afc789b3ba.
Thanks though!

PR #1120, opened 2016-10-04
Drop ElasticSearch datastore on clean

awest1339:
When cuckoo.py --clean is run, drop the ElasicSearch indexes associated with the Cuckoo reports.

If the ElasticSearch indexes are not dropped when the clean is performed, then deleted tasks will appear in the Web UI search even though that task / report doesn't exist anymore.

### DIFF
--- a/lib/cuckoo/core/startup.py
+++ b/lib/cuckoo/core/startup.py
@@ -487,6 +487,28 @@ def cuckoo_clean():
         except:
             log.warning("Unable to drop MongoDB database: %s", mdb)
 
+    # Check if ElasticSearch reporting is enabled and drop that if it is.
+    cfg = Config("reporting")
+    if cfg.elasticsearch and cfg.elasticsearch.enabled:
+        from elasticsearch import Elasticsearch
+        hosts = []
+        for host in cfg.elasticsearch.get("hosts", "127.0.0.1:9200").split(","):
+            if host.strip():
+                hosts.append(host.strip())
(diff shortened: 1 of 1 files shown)


lib/cuckoo/core/startup.py
```diff
@@ -487,6 +487,28 @@ def cuckoo_clean():
...
+        cuckoo_index = cfg.elasticsearch.get("index", "cuckoo")
+        try:
+            # Get all indexes stored in ES
+            # Then drop all starting with cuckoo index
+            # prefix (clears previously rotated indexes)
+            es = Elasticsearch(hosts)
+            indexes = es.indices.get_aliases().keys()
+            for index in indexes:
+                if index.startswith("%s-" % cuckoo_index):
+                    es.indices.delete(index=index, ignore=[400, 404])
```
swackhamer: why not just do index aliases?  es.indices.delete(index=index+"-*")

Just wondering why are you dropping aliases as well?  Cuckoo should not configure any of those.

awest1339: No real reason for it, was just using it to get the names of the cuckoo indexes. Updated the commit to use the regex delete you suggested.

jbremer:
Thanks @awest1339! As @swackhamer said he's incorporated this change into his, more complete, PR, so I'll close this issue and eventually (after some additional feedback) end up merging that PR. Thanks for your contribution!

PR #1121, opened 2016-10-05
forward email from smtp_sinkhole to email address

doomedraven:
in case where email server is not in the same network with cuckoo, is
good to be able forward them to specific email address


jbremer:
As I mentioned in your other SMTP-related I think this is something that should be done by `postfix`. Also we should probably deprecate this file as it's no longer relevant. Could probably be replaced with a `curl` call to an API endpoint that does `sflock` - to get back to what I mentioned in the other PR.
Closing as this won't be merged anyway.. sorry ;-)

PR #1123, opened 2016-10-10
fix LATEST_HTTPREPLAY warning in webgui

doomedraven:
>>> import httpreplay
>>> getattr(httpreplay, "__version__", None)
'0.1.15'


jbremer:
All of this has been removed in #863 as we can specify the required version of `httpreplay` in the required dependency list, but for now I'll take it anyway. Thanks!

PR #1124, opened 2016-10-13
Fix errror "AnalysisManager has no attribute route."

consen:
self.unroute_network() may cause error "AnalysisManager has no attribute route." if self.route_network() doesn't run.


jbremer:
Which exact version of Cuckoo are you running? And can you provide a full stacktrace of the exception - I can't seem to find any usage of the `self.route` variable in `scheduler.py`. Thanks.

consen:
Sorry, it is "package" branch, I will create a new pull request.

Issue #759, opened 2016-02-11
Summary log improvement on NtCreateFile Win API


Titotix:
Currently, the (processing of Win API NtCreateFile)[https://github.com/cuckoosandbox/cuckoo/blob/master/modules/processing/platform/windows.py#L219] for the summary does not care about `creationOptions` flag (See (NtCreateFile doc)[https://msdn.microsoft.com/en-us/library/bb432380%28v=vs.85%29.aspx]). According to this flag, the NtCreateFile call can create or open a directory. As now, a created or opened directory will be logged in summary as a file created or opened.
See #709

jbremer:
Did you have any progress on the directory creation / open thing @Titotix?
I guess we could simply swap out `file_created` by `directory_created` if the `CreateOptions` contains `FILE_DIRECTORY_FILE`? :)

jbremer:
Any update on this @Titotix?

Titotix:
I will go back on it in a few, next week or so ! I wasn't in a hacking 
mood this summer :)

Titotix:
Hey =)
The creation of a dir is not a pb to log, it is a simple case.
But we are also able to "open" a directory with `NtCreateFile`. I am wondering, what could be relevant for us to log about opening a dir ? actually, open a dir with `NtCreateFile` provide a directory handler which serves on other few sys call. [See MS doc](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365258%28v=vs.85%29.aspx). It is mainly about getting some meta data about the dir.
As we are talking about summary info, perhaps it is not relevant to log the opened action but the "following" one, those listed in the link I provided (at least the relevant one) ?

PR #1126, opened 2016-10-13
improve summary info for NtCreateFile #759

Titotix:
See #759.
This PR improve granularity for NtCreateFile about directory manipulation.
It also includes some missing summary info about files/regkey/directory in the web interface but those changes will be outdated when `package` (#1070) branch will be merged. (as it is dynamically created in this branch, not statically as here)


jbremer:
We don't have the "summary listings" anymore, but I guess we should take another look at the return value of `NtCreateFile` some other time.

PR #1129, opened 2016-10-14
Fix errror "AnalysisManager has no attribute route."

consen:
self.unroute_network() may cause error "AnalysisManager has no attribute route." if self.route_network() doesn't run.


jbremer:
I guess this only happens if the analysis didn't run successfully, but sure, I'll take it. Thanks!

PR #1131, opened 2016-10-15
Improved configuration parameter type checking

iN3O:
Added Configuration Parameter Type checking for INTEGER , STRING , PATH , BOOLEAN and UUID.


jbremer:
I've pushed the `strict-config` branch which includes this PR as well as some additional commits. There are a couple of things that still need to happen before we can finally merge it.
* Integrate configuration checking in `startup.py` (i.e., the fields of all sections have to be checked if they *don't* have an `enabled` field or if the `enabled` field is enabled).
* More unit tests to ensure that Cuckoo doesn't throw odd exceptions with this new functionality.
* Type checking for IP addresses, addition of `IntRange` to specify a range for, e.g., `timeout` fields.

PR #1132, opened 2016-10-19
Make it possible to set the snapshot intervals dynamically

knqyf263:
The interval to take screenshots is hard-coded (SHOT_DELAY).
https://github.com/cuckoosandbox/cuckoo/blob/master/analyzer/windows/modules/auxiliary/screenshots.py#L16

So, I made it possible to set SHOT_DELAY from WebUI.

### DIFF
--- a/analyzer/android/modules/auxiliary/screenshots.py
+++ b/analyzer/android/modules/auxiliary/screenshots.py
@@ -13,7 +13,6 @@ from lib.api.adb import take_screenshot
 from lib.api.screenshot import Screenshot
 
 log = logging.getLogger(__name__)
-SHOT_DELAY = 2
 
 class Screenshots(Auxiliary, Thread):
     """Take screenshots."""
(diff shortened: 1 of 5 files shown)


knqyf263:
Thank you for your reply!
For example, when I analyze a standard malware, I set `timeout` to 120 seconds.
In that case, it's ok that the screenshot interval is 1 second.

However, there is malware which wait for a long time after startup.
Ursnif malware which is popular in Japan wait for about 10 minutes after being executed.
So I set `timeout` to 1800 seconds.
If the interval of screenshot is 1 second, 1800 screenshots are taken.
I'd like to set interval to 30 seconds.

As mentioned above, I want to change according to the type of malware (every analysis).

jbremer:
What we should actually improve is the duplication check so that screenshots that look similar are more properly ignored.

knqyf263:
That makes sense. It seems to be difficult to check that screenshots are similar if the simulated human interaction is enabled or malware like ransomware creates a lot of files in desktop, etc.

I sometimes analyze the same malware again because malware communicates with new domain names in a few days. In this case, I want to set a long interval.

Even if the duplication check is improved, I think it is better that the interval can be changed.

jbremer:
Alright, fair enough. This looks fine to me for merging, but I'll wait a little bit for that - will do it after our upcoming release ;-)

knqyf263:
Thanks for your reply.
I'm looking forward to it!

PR #1142, opened 2016-10-27
Handle a condition of connection failure.

chehsunliu:
Handle a condition that connecting to http://api.cuckoosandbox.org/checkversion.php failed.


jbremer:
Thanks!

PR #1169, opened 2016-11-14
resultserver uses threadpool, fix "Too many open files" and "can't start new thread".

consen:
The ResultServer now uses SocketServer.ThreadingTCPServer, which will create a separate thread to handle each request, if dropped files are too many, may cause "Too many open files" or "can't start new thread" error.

ResultServer uses threadpool can limit the number of threads.

### DIFF
--- a/cuckoo/core/resultserver.py
+++ b/cuckoo/core/resultserver.py
@@ -6,12 +6,14 @@
 import errno
 import json
 import os
+import sys
 import socket
 import select
 import logging
 import datetime
 import SocketServer
 import threading
+import Queue
 
 from cuckoo.common.abstracts import ProtocolHandler
 from cuckoo.common.config import Config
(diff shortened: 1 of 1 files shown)


jbremer:
So if ten files are uploaded in parallel, the rest breaks / freezes? ;-)
(I think) I'd prefer having an optionally separate resultserver based on `gevent` that takes as many greenlets (rather than threads) as required.

consen:
@jbremer The rest will block, until a free slot is available. I have tested threadpool, it worked well and no delay too much.

alanyang3:
@consen I'd recommend taking a look at [PR #1170 ](https://github.com/cuckoosandbox/cuckoo/pull/1170) in addition to your changes. I found a number of thread/fd leaks that were contributing to the high thread usage in the resultserver

jbremer:
Yeah, I'm sorry, but this change is definitely not going to be included, although I would like to thank you for your effort :-) There are certainly improvements to be made for analyses where 1000+ files are modified due to ransomware, but throttling 'em with 10 threads at a time isn't going to cut it.

consen:
This PR is not perfect, when malware starts more than 10 processes, the sockets will be exhausted because of log server, leading to the drop files can't dump.

jbremer:
Exactly ;-)

PR #1170, opened 2016-11-15
Cleans up fd usage in resultserver to prevent fd and thread leaks

alanyang3:


jbremer:
Thanks! Looks like there are some goodies in here, but I'll have to double-check them later!

alanyang3:
This PR has been open for over a year. Are these changes still valid? If not, then I will close the PR.

alanyang3:
Closing as this is unlikely valid after two years

PR #1175, opened 2016-11-21
Cuckoomx

doomedraven:
* Allows receive samples by email, unpack them and add to cuckoo
* Send notification when sample reported to the sender(by from field)
* If sample is in cuckoomx database it will just send the report link for hash

### DIFF
--- a/conf/cuckoomx.conf
+++ b/conf/cuckoomx.conf
@@ -1,0 +1,9 @@
+[cuckoomx]
+server = 
+port = 465
+user = 
+password =
+db = sqlite:///cuckoomx.db
+archive_password = infected
+cuckoo_url = http://host:8000/analysis/
+
--- a/cuckoomx.py
+++ b/cuckoomx.py
@@ -1,0 +1,154 @@
+#!/bin/python
+import os
+import sys
+import email
+import imaplib
+import hashlib
+import tempfile
+import smtplib
+from time import sleep
+import mimetypes
+from email import Encoders
+from email.mime.text import MIMEText
+from email.MIMEMultipart import MIMEMultipart
+from email.MIMEBase import MIMEBase
+
+try:
+    from sflock.main import unpack
+    from sflock.unpack import ZipFile
+    from sflock.abstracts import File
+except ImportError:
+    print "Missed deps"\
+    "sudo apt-get install p7zip-full rar unace-nonfree"\
+    "sudo pip install -U sflock"
+
+# Cuckoo root
+CUCKOO_ROOT = os.path.join(os.path.abspath(os.path.dirname(__file__)), "..")
+sys.path.append(CUCKOO_ROOT)
+from lib.cuckoo.common.config import Config
+email_config = Config("cuckoomx")
+
+from lib.cuckoo.core.database import Database
+from lib.cuckoo.common.utils import store_temp_file
+
+from sqlalchemy.sql import func
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import relationship
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy import Column, Integer, String, Boolean, DateTime
+
+from apscheduler.schedulers.background import BackgroundScheduler
+main_db = Database()
+Base = declarative_base()
+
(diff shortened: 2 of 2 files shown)


Nwinternights:
Hi @doomedraven , Cuckoomx is a great tool  but Im stuck with an error when a particular smtp address send an email:
example 
john doe <john.doe@domain.com>

this is the output I get:
send: 'mail FROM:<> size=1402604\r\n'
reply: '250 2.1.0 Sender OK\r\n'
reply: retcode (250); Msg: 2.1.0 Sender OK
send: u'rcpt TO:<doe>\r\n'
reply: '501 5.1.3 Invalid address\r\n'
reply: retcode (501); Msg: 5.1.3 Invalid address
send: u'rcpt TO:<John>\r\n'
reply: '501 5.1.3 Invalid address\r\n'
reply: retcode (501); Msg: 5.1.3 Invalid address
send: 'rset\r\n'
reply: '250 2.0.0 Resetting\r\n'
reply: retcode (250); Msg: 2.0.0 Resetting
{u'"Doe': (501, '5.1.3 Invalid address'), u'John" <John.doe@domain.com>': (501, '5.1.3 Invalid address')}
do you have any idea?
thanks a lot 
M.

doomedraven:
looks like TO not specified in correct way, so no receptors, how you sent the email?

doomedraven:
ah wait that is response with analysis link no?

Nwinternights:
Yep, Instead of URL Link  I just trying to send the pdf report as attachment. With my  mail I've no problem (the reply mail arrives without problem). I get problem with parsing the "TO" from a particular smtp sender.

Nwinternights:
probably because the sender is like: john, Doe(with comma after the name) <john.doe @ domain.com> (without whitespaces)

doomedraven:
yes i think coma can break it, `{u'"Doe': (501, '5.1.3 Invalid address'), u'John" John.doe@domain.com': (501, '5.1.3 Invalid address')}` <- as you can see it not parsed correctly, but it use default parser, i looking at code to see if we can trick it

Nwinternights:
I really appreciated, meanwhile I try to solve it too.
Tks a lot  
M.

doomedraven:
we probably could trick it, but easier can is change just name as it looks like it not follow, RFC822 standart

```
typ, data = conn.fetch(num,"(RFC822)")
msg = email.message_from_string(data[0][1])
```

Nwinternights:
You're right but Im afraid that It can't be possible to change how is build the smtp description of that person because it's an MS Exchange--AD infrastructure so all Infos are stored on the Active Directory. 
It has to be changed there but I guess it's almost impossible :-)
by the way if you find the trick i'll be so glad .
best regards
M.

doomedraven:
ok good to know, then i need trick it in cuckoomx side, i will get a look, i hope today a bit later

Nwinternights:
@doomedraven, I think I've found the solution by using Regex.
import re
#server.sendmail(email_config.cuckoomx.get("from"), to.split(", "), msg.as_string())
server.sendmail(email_config.cuckoomx.get("from"), re.findall(r'[\w\.-]+@[\w\.-]+', to), msg.as_string())

regards
M.

doomedraven:
@Nwinternights thanks, changed :)

jbremer:
I like the idea, but it needs some work. First of all, forwarding email should be a `postfix` configuration setting rather than a Cuckoo feature (some docs on that would be great, though). Secondly, since this PR was created quite some `sflock` work has been done and I think I'd prefer to see all of the email parsing to be done exclusively by `sflock`. That said, this PR would be a good excuse to get started on an automated variant of `sflock` integration in Cuckoo, i.e., also adding an endpoint in the Cuckoo API such as `/tasks/create/archive` or whatever where users can submit `.zip`'s with dozens or hundreds of files.

doomedraven:
make sense, i will add it once i will have a bit more of spare time :)

Roucuck:
can you help me please i add cuckoomx.py and cuckoomx.conf  to my configuration but it doesen't work anyhelp pleasee :/

doomedraven:
no works, doesn't say nothing, logs? proof? anything?

PR #1181, opened 2016-11-22
Core feedback

skftn:
This pull request contains not only the feedback class migrated to Cuckoo core:

- Started working on an abstract mongo class that gives easier access to nested keys (cleaner code). This is not unit-tested yet.
- Every render_template() has been replaced by a custom render() that injects environment variables into each rendered view, such as "page_title" and "page_category", among other things. This was needed for the new sidebar.
- Small changes in error handling for specific pages (return our lovely error page with a correct html status code)
- I got rid of all the relative imports within Cuckoo web and replaced them with full absolute imports
- Changes to `config.py` from core are temporarily; as Config() was revamped by @jbremer in another branch (afaik) that has not been merged yet. When that branch gets merged, the code to interact with the `Config` class in the feedback class should be modified; at which point the unit tests come in handy to validate the changes.
- The feedback class is fully unit tested (apart from URL analysis), however you should check to see if you find `test_feedback.py` sufficient.


jbremer:
Thanks!

PR #1188, opened 2016-11-26
LibVirt machinery should create the KVM domain

sempervictus:
Currently, LibVirt machinery can determine the snapshot and revert
the VM state to it, but it jumps straight to waiting for a state
change without actually issuing a create call to the domain being
used.

Add a single line to call self.vms[label].create() after resolving
the snapshot state and prior to entering the wait state.

### DIFF
--- a/lib/cuckoo/common/abstracts.py
+++ b/lib/cuckoo/common/abstracts.py
@@ -405,7 +405,7 @@ class LibVirtMachinery(Machinery):
             self._disconnect(conn)
             raise CuckooMachineError("No snapshot found for virtual machine "
                                      "{0}".format(label))
-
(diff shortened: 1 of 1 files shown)


jbremer:
@doomedraven @mak thoughts?

doomedraven:
it sounds like vm is just not registered use case no? because it works fine for me

mak:
not sure but you only need to do `create` if you snapshot is not in running state

jbremer:
Closing PR. Doesn't seem to be solving anything and nobody else jumped in to +1 it. Thanks, though!

sempervictus:
No worries, works well for our use, already in our Arch packaging. Thanks for review though.

PR #1195, opened 2016-12-01
Elasticsearch 5.0 compatibility and bug fixes

swackhamer:
Compatibility with Elasticsearch 5.0.1 python module.  This has also been tested with ES 2.x and ES 5.x.

Turned date detection off in the field mapping.  This solves an issue that @keithjjones was having where yara rules had a valid date in a meta attribute that was getting automatically parsed by Elasticsearch as a date field.  The default behavior now is to turn off date detection except when explicitly stated in the elasticsearch json.

This also incorporates PR #1120 as well as dropping the cuckoo template with the cuckoo indexes.

Fixed signatures which sometimes had a key value structure where a value was sometimes a object (dictionary) and sometimes a string.  Now dictionaries are explicitly named <key>_dict.

Also cleaned up parts of the elasticsearch reporting module code.  Due to the new elasticsearch mapping for calls we do not need the process_call function anymore as it will be taken care of by the mapping.

ES 5.0 server fix was to change es.create to es.index as now you need to specify a id if you are issuing a create in ES 5.0.

Increased timeout to index documents from the default 10 seconds to 5 minutes.  This can also be changed with a "hidden" timeout option in the reporting.conf file.

Many many thanks to @keithjjones who worked with me to find the yara date detection bug!


keithjjones:
👍 Thanks!

keithjjones:
This is the error I have currently with your esdev branch after --clean and wiping out index/templates:

```
WARNING:elasticsearch:POST http://127.0.0.1:9200/cuckoo-2016/cuckoo [status:400 request:7.032s]
WARNING:lib.cuckoo.core.plugins:The reporting module "ElasticSearch" returned the following error: Failed to save results in ElasticSearch for task #1: TransportError(400, u'illegal_argument_exception', u'[AQeRIQH][10.0.2.2:9300][indices:data/write/index[p]]')
ERROR:lib.cuckoo.core.plugins:Failed to run the reporting module "MongoDB":
  File "/Source/cuckoo-esdev/utils/../lib/cuckoo/core/plugins.py", line 533, in process
    current.run(self.results)
  File "/Source/cuckoo-esdev/utils/../modules/reporting/mongodb.py", line 253, in run
    self.db.analysis.save(report)
  File "/Source/Python2/lib/python2.7/site-packages/pymongo/collection.py", line 1903, in save
    check_keys, manipulate, write_concern)
  File "/Source/Python2/lib/python2.7/site-packages/pymongo/collection.py", line 430, in _insert
    gen(), check_keys, self.codec_options, sock_info)
DocumentTooLarge: BSON document too large (35522621 bytes) - the connected server supports BSON document sizes up to 16777216 bytes.
```

Issue #1178, opened 2016-11-21
Elasticsearch v5 Error WIth Cuckoov2-RC2?


keithjjones:
I have a fresh install of Cuckoo v2-RC2 on a mac working really well.  I tried to tie it to Elasticsearch through brew (v5 is now the current version there).  I'm getting these errors and no data is being saved to ES.  Any ideas?  Is this a bug?

```
2016-11-21 14:28:13,279 [elasticsearch] WARNING: POST /cuckoov2-1-2016/cuckoo?op_type=create [status:400 request:0.026s]
2016-11-21 14:28:13,294 [lib.cuckoo.core.plugins] WARNING: The reporting module "ElasticSearch" returned the following error: Failed to save results in ElasticSearch for task #1: TransportError(400, u'action_request_validation_exception', u'Validation Failed: 1: an id must be provided if version type or value are set;')
```

That's all it says in cuckoo.log.

This is the relevant reporting.conf information I am using:

```
[elasticsearch]
enabled = yes
# Comma-separated list of ElasticSearch hosts. Format is IP:PORT, if port is
# missing the default port is used.
# Example: hosts = 127.0.0.1:9200, 192.168.1.1:80
hosts = 127.0.0.1:9200
# Set to yes if we want to be able to search every API call instead of just
# through the behavioral summary.
calls = no
# Index of this Cuckoo instance. If multiple Cuckoo instances connect to the
# same ElasticSearch host then this index (in Moloch called "instance") should
# be unique for each Cuckoo instance.
#
index = cuckoov2-1
#
# Logging time pattern.  This sets how elasticsearch creates indexes
# by default it is yearly in most instances this will be sufficient
# valid options: yearly, monthly, daily
#
index_time_pattern = yearly
#
# Cuckoo node name in Elasticsearch to identify reporting host. Can be useful
# for automation and while referring back to correct Cuckoo host. 
#
# cuckoo_node = "cuckoo.example.host"
```

keithjjones:
Well, it's not ElasticSearch on a mac that is causing an issue.  I stood up an ElasticSearch instance on a CentOS computer and I'm getting the same type of error:

```
2016-11-21 16:09:13,861 [elasticsearch] WARNING: HEAD /_template/cuckoo_template [status:404 request:0.016s]
2016-11-21 16:09:15,294 [elasticsearch] WARNING: POST /cuckoov2-1-2016/cuckoo?op_type=create [status:400 request:0.771s]
[...]

Issue #1183, opened 2016-11-23
cuckoo.py --clean should clean the ES instance too


keithjjones:
If not, old data is left in the ES server.  Right now, it has to be cleaned by hand with a:

```
curl -X DELETE ESHOST:9200/cuckoo-2016 
```

swackhamer:
https://github.com/cuckoosandbox/cuckoo/pull/1120 is a PR that provides that functionality.

jbremer:
Will close this issue as it's being handled as part of #1196 (which is superior to #1120). Merging that PR hopefully sooner rather than later. Thanks!

PR #1196, opened 2016-12-01
Elasticsearch 5.0 client library compatibility and bug fixes

swackhamer:
Compatibility with Elasticsearch 5.0.1 python module. This has also been tested with ES 2.x and ES 5.x.

Fixes bugs: https://github.com/cuckoosandbox/cuckoo/issues/1178 https://github.com/cuckoosandbox/cuckoo/issues/1183

Turned date detection off in the field mapping. This solves an issue that @keithjjones was having where yara rules had a valid date in a meta attribute that was getting automatically parsed by Elasticsearch as a date field. The default behavior now is to turn off date detection except when explicitly stated in the elasticsearch json.

This also incorporates PR #1120 as well as dropping the cuckoo template with the cuckoo indexes.

Fixed signatures which sometimes had a key value structure where a value was sometimes a object (dictionary) and sometimes a string. Now dictionaries are explicitly named _dict.

Also cleaned up parts of the elasticsearch reporting module code. Due to the new elasticsearch mapping for calls we do not need the process_call function anymore as it will be taken care of by the mapping.

ES 5.0 server fix was to change es.create to es.index as now you need to specify a id if you are issuing a create in ES 5.0.

Increased timeout to index documents from the default 10 seconds to 5 minutes. This can also be changed with a "hidden" timeout option in the reporting.conf file.

Many many thanks to @keithjjones who worked with me to find the yara date detection bug!

### DIFF
--- a/data/elasticsearch/template.json
+++ b/data/elasticsearch/template.json
@@ -12,7 +12,49 @@
         "cuckoo": {
             "dynamic_templates": [
                 {
-                    "notanalyzed": {
+                    "not_analyzed": {
+                        "mapping": {
+                            "index": "not_analyzed",
+                            "type": "string"
+                        },
+                        "match_mapping_type": "string",
+                        "match": "*"
+                    }
+                },
+                {
+                    "signatures": {
+                        "mapping": {
+                            "index": "not_analyzed",
+                            "type": "string"
+                        },
+                        "path_match": "signatures.marks.call.arguments.*",
+                        "path_unmatch": "signatures.marks.call.arguments.registers.*",
+                        "match": "*",
+                        "match_mapping_type": "long"
+                    }
+                }
+            ],
+            "properties": {
+                "report_time": {
+                    "format": "epoch_second",
+                    "type": "date"
+                },
+                "procmemory": {
+                    "type": "nested",
+                    "include_in_root": "True",
+                    "properties": {
+                        "regions": {
+                           "include_in_root": "True",
+                           "type": "nested"
+                        }
+                    }
+                }
+            }
+        },
+        "call": {
+            "dynamic_templates": [
+                {
+                    "not_analyzed": {
                         "mapping": {
                             "index": "not_analyzed",
                             "type": "string",
@@ -21,26 +63,16 @@
                         "match_mapping_type": "string",
                         "match": "*"
                     }
-                }
-            ],
-            "properties": {
-                "report_time": {
-                    "format": "epoch_second",
-                    "type": "date"
-                }
-            }
-        },
-        "call": {
-            "dynamic_templates": [
+                },
                 {
-                    "notanalyzed": {
+                    "call_arguments": {
                         "mapping": {
                             "index": "not_analyzed",
-                            "type": "string",
-                            "doc_values": "True"
+                            "type": "string"
                         },
-                        "match_mapping_type": "string",
-                        "match": "*"
+                        "path_match": "arguments.*",
+                        "match": "*",
+                        "match_mapping_type": "long"
                     }
                 }
             ],
--- a/data/elasticsearch/template.json
+++ b/data/elasticsearch/template.json
@@ -1,5 +1,4 @@
 {
-    "order": 0,
     "template": "cuckoo-*",
     "settings": {
         "index": {
@@ -11,6 +10,7 @@
     },
     "mappings": {
         "cuckoo": {
+            "date_detection": false,
             "dynamic_templates": [
                 {
                     "not_analyzed": {
@@ -27,7 +27,8 @@
                     "signatures": {
                         "mapping": {
                             "index": "not_analyzed",
-                            "type": "string"
+                            "type": "string",
+                            "ignore_above": 32766
                         },
                         "path_match": "signatures.marks.call.arguments.*",
                         "path_unmatch": "signatures.marks.call.arguments.registers.*",
@@ -54,13 +55,14 @@
             }
         },
         "call": {
+            "date_detection": false,
             "dynamic_templates": [
                 {
                     "not_analyzed": {
                         "mapping": {
                             "index": "not_analyzed",
                             "type": "string",
-                            "doc_values": "True"
+                            "ignore_above": 32766
                         },
                         "match_mapping_type": "string",
                         "match": "*"
@@ -70,7 +72,8 @@
                     "call_arguments": {
                         "mapping": {
                             "index": "not_analyzed",
-                            "type": "string"
+                            "type": "string",
+                            "ignore_above": 32766
                         },
                         "path_match": "arguments.*",
                         "match": "*",
@@ -85,6 +88,5 @@
                 }
             }
         }
-    },
-    "aliases": {}
+    }
 }
--- a/data/elasticsearch/template.json
+++ b/data/elasticsearch/template.json
@@ -16,7 +16,8 @@
                     "not_analyzed": {
                         "mapping": {
                             "index": "not_analyzed",
-                            "type": "string"
+                            "type": "string",
+                            "ignore_above": 32766
                         },
                         "match_mapping_type": "string",
                         "match": "*"
--- a/data/elasticsearch/template.json
+++ b/data/elasticsearch/template.json
@@ -3,6 +3,7 @@
     "template": "cuckoo-*",
     "settings": {
         "index": {
+            "mapping.coerce": true,
             "number_of_shards": "1",
             "codec": "best_compression",
             "number_of_replicas": "0"
--- a/lib/cuckoo/core/startup.py
+++ b/lib/cuckoo/core/startup.py
@@ -488,6 +488,29 @@ def cuckoo_clean():
         except:
             log.warning("Unable to drop MongoDB database: %s", mdb)
 
+    # Check if ElasticSearch reporting is enabled and drop that if it is.
+    cfg = Config("reporting")
+    if cfg.elasticsearch and cfg.elasticsearch.enabled:
+        from elasticsearch import Elasticsearch
+        hosts = []
+        for host in cfg.elasticsearch.get("hosts", "127.0.0.1:9200").split(","):
+            if host.strip():
+                hosts.append(host.strip())
+        cuckoo_index = cfg.elasticsearch.get("index", "cuckoo")
+        try:
+            # Get all indexes stored in ES
+            # Then drop all starting with cuckoo index
+            # prefix (clears previously rotated indexes)
+            es = Elasticsearch(hosts)
+            es.indices.delete(index='%s-*' % cuckoo_index, ignore=[400, 404])
+
+            cuckoo_template = cuckoo_index + "_template"
+            if es.indices.exists_template(cuckoo_template):
+                # also drop the dynamic template
+                es.indices.delete_template(cuckoo_template)
+        except:
+            log.warning("Unable to drop ElasticSearch database: %s", index)
+
     # Paths to clean.
     paths = [
         os.path.join(CUCKOO_ROOT, "db"),
--- a/lib/cuckoo/core/startup.py
+++ b/lib/cuckoo/core/startup.py
@@ -509,7 +509,7 @@ def cuckoo_clean():
                 # also drop the dynamic template
                 es.indices.delete_template(cuckoo_template)
         except:
-            log.warning("Unable to drop ElasticSearch database: %s", index)
+            log.warning("Unable to drop ElasticSearch database %s", cuckoo_index)
 
     # Paths to clean.
     paths = [
--- a/modules/reporting/elasticsearch.py
+++ b/modules/reporting/elasticsearch.py
@@ -107,9 +107,11 @@ class ElasticSearch(Report):
             # template.json.
             cuckoo_template["template"] = self.index + "-*"
 
-        self.es.indices.put_template(
-            name=self.template_name, body=json.dumps(cuckoo_template)
-        )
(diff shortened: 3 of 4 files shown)


keithjjones:
👍

zippy1990:
Been testing this PR when merged into my Cuckoo 2.0-RC1 with ES 5.0.1 but I still received the following error:

`WARNING:lib.cuckoo.core.plugins:The reporting module "ElasticSearch" returned the following error: Failed to save results in ElasticSearch for task #3: TransportError(400, u'illegal_argument_exception', u'[ZgC-WzA][127.0.0.1:9300][indices:data/write/index[p]]')
`

Which causes to leave that specific task out of the ES results. Out of 5 analysis results, only 1 gave the error.

jbremer:
@zippy1990 Did you create the ES database *after* including this PR into your setup? Also, you might want to upgrade to `2.0-rc2`.

zippy1990:
@jbremer I cleaned the ES database before incorporating the PR but I guess that doesn't suffice. Will try as suggested and report here.

zippy1990:
Created a new Cuckoo installation from scratch with RC2 and deleted the ES index for Cuckoo. The previous error is gone, got this one instead:

`WARNING:lib.cuckoo.core.plugins:The reporting module "ElasticSearch" returned the following error: Failed to save results in ElasticSearch for task #4: create() takes at least 5 arguments (5 given)
`
Because sure, that makes so much sense. You got your 5 arguments create(), what else do you want from me.

keithjjones:
@zippy1990 be sure to delete the template too.  --clean might do that.  @swackhamer can advise.

@jbremer and @zippy1990 I find about 20% of my analyses fail in 2.4 or 5.0.  Most of the time, it is because the json is too big for mongodb.  Plus, there still seems to be something inserted into ES that ES doesn't like regardless of version.  This PR, as far as I know, gets 5.0 working like 2.4.  This looks like the same issue I'm having that probably needs to be addressed with a separate fix.

zippy1990:
@keithjjones Yeah, I've set up a Cuckoo 2.0 RC2 installation just now, incorporated this PR and deleted my previous ES indexes but  occasionally (1 in 10 analyses or so) get the error mentioned in my first comment.

keithjjones:
@zippy1990 that sounds like the failure rate I'm seeing as well.

swackhamer:
I have not been able to diagnose this issue that you are getting zippy1990 with ES 5.0.  Keithjjones is also getting "indices:data/write/index[p]]" and we debugged it so far as to drop the json document out to a file before it got indexed by ES.  That same document that got dropped out would index into ES fine with no issues...

We even attempted to turn the Cuckoo report into JSON using mongo's "bson.json_util.dumps" http://api.mongodb.com/python/current/api/bson/json_util.html to turn it into a JSON document before the Cuckoo report got into ES but that did not work...

Normally with ES it gives you a better error then just "indices:data/write/index[p]", if it was a mapping issue or some other issue it will normally tell you the field name that is having issues.

The one thing that should not affect this in any way is that we need to maintain two mappings.  1 or ES 2.X and one for ES 5.X as the "string" type has been deprecated.  https://www.elastic.co/guide/en/elasticsearch/reference/current/string.html

Keith was having issues with ransomware that was creating extraordinarily large Cuckoo reports (1+GB).  Are your reports also very large?

zippy1990:
@swackhamer So I just did another test run with same setup as before but with a different set of files (ranging from zip, exe to pdf and word documents) and for the 15 analysis cases I don't seem to get very large reports. Totaling the 15 analysis folders results in about 90mb of disk usage, which is nowhere near the 1gb that you mentioned for 1 report. 

Striking to me was that this time around I didn't get the error mentioned before (indices:data/write/index[p]) but instead got the 2 errors shown below (both once during the analysis of the 15 items). Please note that I am using the dedicated reporting module in auto mode with a Postgresql Database, I'm not sure if the issues would occur if used with the build in reporting module.

`ERROR:sqlalchemy.pool.QueuePool:Exception during reset or similar
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/pool.py", line 636, in _finalize_fairy
    fairy._reset(pool)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/pool.py", line 774, in _reset
    self._reset_agent.rollback()
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/base.py", line 1563, in rollback
    self._do_rollback()
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/base.py", line 1601, in _do_rollback
    self.connection._rollback_impl()
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/base.py", line 670, in _rollback_impl
    self._handle_dbapi_exception(e, None, None, None, None)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/base.py", line 1341, in _handle_dbapi_exception
    exc_info
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/util/compat.py", line 199, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/base.py", line 668, in _rollback_impl
    self.engine.dialect.do_rollback(self.connection)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/default.py", line 420, in do_rollback
    dbapi_connection.rollback()
DatabaseError: (psycopg2.DatabaseError) error with status PGRES_TUPLES_OK and no message from the libpq
`

I've seen that one before and I'm not even sure if it's related to ES in any way, tho I only noticed it after adding ES to the environment. The other exception that I got is:

`ERROR:lib.cuckoo.core.plugins:Failed to run the processing module "Static" for task #9:
  File "[redactedFilePath]/cuckoo/utils/../lib/cuckoo/core/plugins.py", line 242, in process
    data = current.run()
  File "[redactedFilePath]/cuckoomasterRC2/cuckoo/utils/../modules/processing/static.py", line 695, in run
    pdf_worker, (self.file_path,), timeout=timeout
  File "[redactedFilePath]/cuckoomasterRC2/cuckoo/utils/../lib/cuckoo/misc.py", line 22, in dispatch
    p.start()
  File "/usr/lib/python2.7/multiprocessing/process.py", line 124, in start
    'daemonic processes are not allowed to have children'
AssertionError: daemonic processes are not allowed to have children`

I've tried grepping trough my Elasticsearch logs as suggested in /var/log/elasticsearch/*.log with the keywords of the mentioned exception, but to no avail. The only debugging messages in there are about exceeding the disk storage threshold.

keithjjones:
@zippy1990 I haven't received those errors.  They are new to me.

swackhamer:
@zippy1990 those errors do not look like ES.  The first error looks to be related to the Postgres database based off this line:

`DatabaseError: (psycopg2.DatabaseError) error with status PGRES_TUPLES_OK and no message from the libpq`

Some potential information about that here: https://github.com/psycopg/psycopg2/issues/281

The second error seems to be related to this line: https://github.com/cuckoosandbox/cuckoo/commit/8b51c874bc945a7cb5ea98df4ac0bfbad0d5daff#diff-033aeda7c00b458591305630264df6d3R694 and https://github.com/cuckoosandbox/cuckoo/commit/8b51c874bc945a7cb5ea98df4ac0bfbad0d5daff#diff-093dd89808cc9806aed0e3acf310a92b

It looks like you are using postgres and using process2.py?  That looks like a recently implemented bug and another bug report should probably be opened for it.


Elasticsearch 5.0 has changed a lot more then I originally thought.

It looks like we are going to have to maintain two mappings -- one for ES 2.X and one for ES 5.X.  I don't think the community is "ready" to completely switch over to ES 5.X yet so I think this will be a good solution for a while.

For example string fields have been completely replaced with "keyword" or "text".  For the Cuckoo use case most of the fields will need to be "keyword" as we don't want to break up (most) fields using a string analyzer as it does not make sense to have a domain break into www google and com.  https://www.elastic.co/guide/en/elasticsearch/reference/5.x/keyword.html . I am also experimenting with using eager "Global Ordinals" to speed up aggregations on fields that commonly have similar data...  https://www.elastic.co/guide/en/elasticsearch/reference/5.x/fielddata.html#global-ordinals

Hoping to get something out soon for ES 5.0.  Has anyone had any crashes with this on ES 2.x?

keithjjones:
Is it possible to at least get the --clean functionality into the next 2.0 RC?  Even if you are on 2.4, that's incredibly annoying having to clean it out by hand.

zippy1990:
@swackhamer The problem is that those specific errors only started appearing after using ES in general, also still using process.py (not 2). To answer your question at the end, I got the same errors using 2.x. 

Regarding the two mapping solution, what's the effectiveness of this fix? Is it a temporary one that gives a stable experience at least?

swackhamer:
@zippy1990 I think you may have incorporated in this update https://github.com/cuckoosandbox/cuckoo/commit/8b51c874bc945a7cb5ea98df4ac0bfbad0d5daff#diff-033aeda7c00b458591305630264df6d3R694 when you upgraded to this branch, but it is not necessarily this PR that caused them.  Try turning off the ES reporting module and see if they still persist.  If they still persist then it is not ES, but if they go away it probably is.  You could try to rerun process.py for that specific report after you turn ES reporting off in config.ini to test.  If it is ES I will need to get your analysis folder for that task from the storage directory to debug through.

ES 5.0 is going to need those mapping changes, but ES 2.X *should* work as is.  The reason is they changed how string fields are handled in ES 5.0.  I also added in a couple of "coercion" and corner case protection into this new ES template.  Specifically if we get fields that are over the max ES size and https://github.com/swackhamer/cuckoo/blob/0abc2daeec4d6e5c32bfe9954024ee55751ea3ef/data/elasticsearch/template.json#L5 where coerce is turned on to try to get fields mapped in where they have been previously mapped one way and now are a different type.  Such as if we map in a variable as a "string" and then later try to index that same variable that is a "int".  Coercision will automatically try to index it as a string instead of throwing a mapping parsing error.  https://www.elastic.co/guide/en/elasticsearch/reference/current/coerce.html

This mapping problem specifically has been the issue in Cuckoo's ES implementation.  You can see this with signatures and calls where sometimes a signature's key will have a value that is a dictionary and then later it is a string.  I tried to fix this case in this PR by making all dictionaries mapped in with the appended key (or fieldname) "_dict".  This fixes a bug in mainline Cuckoo right now with some signatures.  This is also now appearing for that #1217 and will have to be handled in a similar way.  Unfortunately ES is not like Mongo where it is COMPLETELY schema less.  ES is schema less for the first document then that document sets the "mappings" or schema for all other documents indexed into that index.  With dynamic templates and dynamic mappings (what template.json sets up) we can control ES's internal behavior.

If no one is having issues with this PR in ES 2.X we should probably merge this and then open a new PR for ES 5.0 SERVER compatibility.  This PR upgrades support ES 5.0 python lib and fixes bugs in ES 2.X.

swackhamer:
I am going to start work on ES 5.0 server compatibility in another PR.  This PR should be good to go for deployment into Cuckoo for ES 5.x client lib and fixes bugs in the current ES 2.x integration.

jbremer:
Hi all, sorry for the late response. I'll be merging this PR soon, but I have a few remarks & questions before doing so.
* Please read up on #863 as I'll be mentioning it a few times here (also referred to as `package` branch). This is the upcoming release and almost all of our development is done there.
* @zippy1990 I've never seen that `PostgreSQL` exception before, if you have any more information on it please do let me know.
* @zippy1990 Interesting. So `process.py` has been deprecated for quite a while as it, among other issues, randomly freezes up. The newer `process2.py` takes care of that in the sense that it doesn't get into odd `multiprocessing` locks (and needs to be started multiple times by the user). In `package` we feature a combination of `process.py` and `process2.py` (i.e., `process2.py` with support for all database backends) and also the user is automatically prompted with a `supervisor` script that does all of starting/stopping `cuckoo process` for him or her. As such `cuckoo process` / `process2.py` is also not affected by your bug as it's not maintained by `multiprocessing` and can spawn as many processes as it likes - which seems to be the problem here.
* In `package` almost all code changes are unit tested, so while merging I'll be doing some of that as well. For this it would be preferable if I could get my hands on a couple of `report.json` files that show *problem* scenarios, e.g., the `Yara` date issue, the `IRMA` PE mapping issue, etc.
* I see that you bumped the `elasticsearch` client API version from `2.2.0` to `5.0.1`, is this still the desired version? As I understand this is compatible with ES `2.4` and `5.x`? I also see ES versions `5.1.0` and `5.2.0` on PyPI now (https://pypi.python.org/pypi/elasticsearch), does that means it's time for an update on this PR?
* You mentioned having separate templates for `2.4` and `5.x` ES, is this still a requirement? If so would you like to have two separate `.json` files or would we be able to use a template (in `jinja2` sense) and replace a couple of fields on-the-fly, e.g., on `2.4` you'd replace some keywords with `"string"` and on `5.x` you'd replace those same keywords with `"keyword"` or whatever is the appropriate ES term (that's one I saw come-by in this thread).
* I like your `_dict` idea, although not the most optimal in terms of beauty I think it's definitely a good idea, so let's keep that for now.
* What is left for `5.x server` support? Is that something you're still planning to work on? :-)

Hopefully addressed all topics in this issue. Thanks again for your contributions and I hope to merge this soon. Again, sorry for the delay ;-)

jbremer:
@swackhamer ping!

jbremer:
Well, I've merged this PR taking into account some of my earlier remarks. Once more users start actually using it, I'm confident we'll get to the desired end result.

PR #1237, opened 2017-01-10
Update human.py with more buttons to click

kevross33:
From spender-sandbox human.py https://github.com/spender-sandbox/cuckoo-modified/blob/master/analyzer/windows/modules/auxiliary/human.py

### DIFF
--- a/analyzer/windows/modules/auxiliary/human.py
+++ b/analyzer/windows/modules/auxiliary/human.py
@@ -50,7 +50,7 @@ def foreach_child(hwnd, lparam):
         "remind me later",
         "save", "sauvegarder"
     ]
-    
+
     # List of buttons labels to not click.
     dontclick = [
         "don't run",
--- a/analyzer/windows/modules/auxiliary/human.py
+++ b/analyzer/windows/modules/auxiliary/human.py
@@ -36,14 +36,21 @@ def foreach_child(hwnd, lparam):
         "agree", "j'accepte",
         "enable", "activer",
         "don't send", "ne pas envoyer",
+        "don't save",
         "continue", "continuer",
         "unzip", "dezip",
         "open", "ouvrir",
+        "close the program",
         "execute", "executer",
         "launch", "lancer",
+        "later",
+        "finish",
+        "end",
+        "allow access",
+        "remind me later",
         "save", "sauvegarder"
     ]
-
+    
     # List of buttons labels to not click.
     dontclick = [
         "don't run",


analyzer/windows/modules/auxiliary/human.py
```diff
@@ -36,14 +36,21 @@ def foreach_child(hwnd, lparam):
...
         "launch", "lancer",
+        "later",
+        "finish",
+        "end",
+        "allow access",
+        "remind me later",
         "save", "sauvegarder"
     ]
-
+
```
jbremer: Please get rid of this small change :)

jbremer:
Looks fine to me, almost ready to merge :)

kevross33:
There you go that is done. 

On another note as I see google summer of code stuff is up I am wondering if I might suggest a project option or section of another project for someone's consideration? I would love to see improved exploit detection in cuckoo and while I would like to be able to implement such a thing myself my ability to code such a thing is no where near that level (the curse of a network background + so much knowledge to learn out there and too little time). 

The reason would be:

- ROP detection (via stack pivot detection) has been implemented by the checkpoint change but ROP may not be used.
- While payload detection is possible (i.e martian processes, dropped files etc) that obviously is on the condition of sucessful exploitation which may not occur because of the usual factors (wrong version, typical exploit sucess rate etc.) Thus earlier detection on the exploit itself may be useful.

Pwnypot was an interesting first iteration a while back but what would be interesting would be:

- Showing key exploit stages in interface (ROP traces, stack traces, shellcode dumping etc.). Being able to see the module being exploited would be interesting (i.e Java, IE, specific office modules) maybe allowing specific exploit to be identified.
- Other ROP detection techniques. ROP detection may even be expanded to Linux and Android (good writup in Android Hacker's Handbook on ROP on ARM even if it is an older book).
- Export and Import Address Table filtering for shellcode detection.
- Control flow - may be difficult in VM though.
- Heap Spray detection (memory growth, tracking what data is being allocated as well for repeating patterns?).
- Implementing Java hooks (specfically setSecurityManager for bypass). I know Java is old but exploits may still work against many machines as where old applications actually use Java rather than it just being installed for the sake of it.

Endgame has done an interesting project with this:

https://www.endgame.com/blog/hunting-exploit-kits
https://github.com/endgameinc/Maxwell

There is code here specific to the exploit detection:

https://github.com/endgameinc/Maxwell/blob/master/flux/hook_inject.cpp

jbremer:
All merged now, thanks again! To answer your questions @kevross33: unfortunately at the time of your writing it was already too late for another GSoC project, however, I do agree that these improvements would be great additional functionality in Cuckoo. Assuming a certain project goes through we will definitely be looking into a few of your proposed aspects, though. So for now.. all I can say is let's wait and see. Regarding the whole exploit analysis improvements; one of the most important parts of that is actually having a good set of exploits that exhibit the behavior as you mentioned. So what I'm saying is, if you have any such exploits and know what environment they should be run in etc, then please do let me know, either here or privately (i.e., `jbr@cuckoo.sh`).
I hope that explains our current situation.. :-)

PR #1239, opened 2017-01-12
Some pathes

doomedraven:
* add `self.route` value to init, as if vm is in an incorrect state, scheduler.py calls unroute_network and it trigger error as self.route not defined and it generates “Failure in AnalysisManager.run”

* QEMU config patch, emu machinery looks for `kernel`, not
`kernel_path`, + added example of arm vm as that one requires `initrd` value

* documentation how to setup them will be in next PR

@jbremer the same persist in package, can you backport it or want me do PR to package?


doomedraven:
FYI https://github.com/cuckoosandbox/cuckoo/blob/master/modules/machinery/qemu.py#L174
there is where it checks for kernel and not kernel_path

```
        # allow some overrides from the vm specific options
        # also do another round of parameter formatting
        for var in ["mac", "kernel", "initrd"]:
            val = getattr(vm_options, var, params.get(var, None))
            if not val:
                continue
            params[var] = val.format(**params)
```

jbremer:
Thanks!

PR #1241, opened 2017-01-15
Edit configuration files from Web User Interface.

tatsui:
Add  web user interface function to edit configuration files.
You need edit this file.
`# vim /etc/sudoers.d/cuckoo`
`cuckoo ALL=NOPASSWD: /sbin/shutdown`
`cuckoo ALL=NOPASSWD: /sbin/reboot`
`cuckoo ALL=NOPASSWD: /sbin/initctl`
`# chown root:root /etc/sudoers.d/cuckoo`
`# chmod 600 /etc/sudoers.d/cuckoo`
cuckoo1
cuckoo2

### DIFF
--- a/web/configure/urls.py
+++ b/web/configure/urls.py
@@ -1,0 +1,13 @@
+# Copyright (C) 2010-2013 Claudio Guarnieri.
+# Copyright (C) 2014-2016 Cuckoo Foundation.
+# This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
+# See the file "docs/LICENSE" for copying permission.
+
+from . import views
+from django.conf.urls import url
+
+urlpatterns = [
+    url(r"^$", views.index),
+    url(r"^machine/add$", views.add),
+    url(r"^machine/delete$", views.delete),
+]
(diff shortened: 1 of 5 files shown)


jbremer:
Excellent work! Unfortunately we've completely reworked the way that we operate on the configuration (not the files, though, they're still the same) as well as various changes to the web interface that would make this hard to merge into #863. This feature is on our todo list though (also as part of other upcoming features), so for now I've asked @automagisch about his thoughts for merging this to the newer web interface.
Thanks for your contribution!

doomedraven:
this is really nice feature, but it would be much better(probably already have, I didn't read the code) if configuration part had some auth, i think with python decorators that should be easy to extend :)

jbremer:
@doomedraven Although a completely different issue/feature, that's also on our todo list, yes ;-)

doomedraven:
@jbremer I mean for config section at least, to have cuckoo open for user, but closed configuration section with access only for admins :)

tatsui:
How about using apache's digest authentication and proxy_pass as a temporary solution?
Only the setting panel (/configure) requests admin authentication.

doomedraven:
Yap as temp solution is totally fine

jbremer:
I much prefer not having such temporary solutions as I'll be the person that's going to have to maintain it.
Due to the fact that it would be easier for us to re-create similar functionality in the new web interface rather than merging over this code based on our older codebase, as well as the fact that it's using some non-default `init` scripts and configuration, and special `sudo` magic that I'd rather not see in any production system, I'm going to close it.
As I mentioned this functionality is on our TODO list, though, so please hang in there while we work on it in the future.

PR #1252, opened 2017-01-22
add the ability to use the authentication process for mongodb

niterain:
We needed authentication to be set for cuckoo to access our MongoDBs, so I added it in.


jbremer:
Hi, thanks for the PR. Interestingly enough we've already implemented this in the `package` branch as we had the same requirement. I did forget to also apply this to the `0.6` database migration though, so thanks for that, I'll be doing that. Although, on the other hand, provided this functionality wasn't around in `0.6`, would it make sense to still add it? Is the authentication database part required?

niterain:
Hi jbremer:

Not sure if I understand what you meant by the functionality wasn't around in 0.6.  Are we talking about when 0.6 came out?  Or do you mean, that people who are on older version of cuckoo, aren't able to upgrade yara, mysql, or mongodb, because there are some checks in cuckoo to prevent newer versions to be used?  I believe authentication has been in mongodb since 2.6, and since the changes are user defined/optional, they have total control if authentication is used or not depending on their setup.

Getting to your other question.  In my case authentication is required, only one database can control authentication to mongodb at our company.  I think this also makes things cleaner as well (you get a nice authentication overview).  If I am reading your code correctly, if I name my cuckoo database "cuckoo_system", I would have to make another database called "cuckoo"  just for authentication purposes?

jbremer:
Users are certainly able to upgrade, but in Cuckoo 0.6 MongoDB authentication wasn't possible/implemented, so I'd somewhat assume that those users wouldn't have authentication enabled in the first place - and in that case it'd be kind of odd to upgrade with authentication, I think?
Regarding `cuckoo_system` vs `cuckoo`, I don't know really. Haven't studied MongoDB that closely yet (or deployed it with authentication, apparently).

jbremer:
For reasons described above I don't think MongoDB authentication is required for the existing database migration. In our upcoming release we're already having MongoDB authentication support though, so I won't be merging this PR :-) Thanks anyway!

PR #1254, opened 2017-01-22
enabled ActiveX controls for office packages. added ability to view strings in web UI for dropped files and buffers.

ronbarrey:
The changes enable ActiveX controls while analyzing Office documents that may contain ActiveX macros. Also added the ability to view strings in the web UI for dropped files and buffers. Moved the buffer download button to match the file download button. Removed unused </h4> tag.


doomedraven:
The duplication of misp integration not make sense, you can better provide mod of current one with ext in config, not in script body

PR #1262, opened 2017-01-26
add suricata check to Signatures class replacement for pr 1062 as i don't have permissions to edit %)

doomedraven:


jbremer:
I'll have to unit test it later, but thanks!

doomedraven:
thanks for merge it, it will make my signatures more clear :)

PR #1279, opened 2017-02-09
Fix KVM sniffing and Debian Testing compatibility  

seanthegeek:
- Update `requirements.txt` to work with OpenSSL 1.1.0c
- Automatically detect and use the `vnetx` interface assigned to KVM on the VM start for sniffing and routing, rather than the `virbr` bridge
- Fix `vpn_status()` in `rooter.py` to properly detect running VPNs in Debian and Ubuntu

### DIFF
--- a/conf/kvm.conf
+++ b/conf/kvm.conf
@@ -15,7 +15,7 @@ interface = virbr0
 label = cuckoo1
 
 # Specify the operating system platform used by current machine
-# [windows/darwin/linux].
+# [windows/darwin/linux]
 platform = windows
 
 # Specify the IP address of the current virtual machine. Make sure that the
--- a/conf/kvm.conf
+++ b/conf/kvm.conf
@@ -29,11 +29,6 @@ ip = 192.168.122.105
 # Example (Snapshot1 is the snapshot name):
 # snapshot = Snapshot1
 
-# (Optional) Specify the name of the network interface that should be used
-# when dumping network traffic from this machine with tcpdump.
-# Example (virbr0 is the interface name):
-# interface = virbr0
-
 # (Optional) Specify the IP of the Result Server, as your virtual machine sees it.
 # The Result Server will always bind to the address and port specified in cuckoo.conf,
 # however you could set up your virtual network to use NAT/PAT, so you can specify here 
--- a/modules/machinery/kvm.py
+++ b/modules/machinery/kvm.py
@@ -27,5 +27,4 @@ class KVM(LibVirtMachinery):
 
     def start(self, label, task):
         super(KVM, self).start(label, task)
-        if not self.db.view_machine_by_label(label).interface:
-            self.db.set_machine_interface(label, self._get_interface(label))
+        self.db.set_machine_interface(label, self._get_interface(label))
--- a/modules/machinery/kvm.py
+++ b/modules/machinery/kvm.py
@@ -3,10 +3,29 @@
 # This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
 # See the file 'docs/LICENSE' for copying permission.
 
+import xml.etree.ElementTree as ET
+
 from lib.cuckoo.common.abstracts import LibVirtMachinery
+
 
 class KVM(LibVirtMachinery):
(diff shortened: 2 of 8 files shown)


utils/rooter.py
```diff
@@ -40,39 +44,57 @@ def rt_available(rt_table):
...
+                running = len(status_file.read()) > 0
+        return running
+
+    config_path = "/etc/openvpn/"
     ret = {}
-    for line in run(settings.service, "openvpn", "status")[0].split("\n"):
-        x = re.search("'(?P<vpn>\\w+)'\\ is\\ (?P<running>not)?", line)
-        if x:
-            ret[x.group("vpn")] = x.group("running") != "not"
+    for vpn in [f for f in listdir(config_path) if os.path.isfile(os.path.join(config_path, f)) and f.endswith(".conf")]:
```
jbremer: Please use something else than list comprehension here ;)

requirements.txt
```diff
@@ -26,11 +26,11 @@ pyasn1==0.1.8
 pycparser==2.14
 pymisp==2.4.54
 pymongo==3.0.3
-pyOpenSSL==0.15.1
+pyOpenSSL==16.2.0
 python-dateutil==2.4.2
 python-editor==0.3
 python-magic==0.4.6
-requests==2.7.0
+requests==2.13.0
```
jbremer: Hold your horses.. before I can merge this some additional unit testing may be required.

jbremer:
@doomedraven @mak thoughts on this PR?

doomedraven:
get iface automatically is a good feature, but what is really fix kvm sniffing? It would need to test the PR, in my setup I don't have any issues with routing/sniffing so probably a bit more of context would be great :)

conf/kvm.conf
```diff
@@ -29,11 +29,6 @@ ip = 192.168.122.105
 # Example (Snapshot1 is the snapshot name):
 # snapshot = Snapshot1
 
-# (Optional) Specify the name of the network interface that should be used
-# when dumping network traffic from this machine with tcpdump.
-# Example (virbr0 is the interface name):
-# interface = virbr0
```
seanthegeek: Correct. This is a port of a fix that I used in `cuckoo-modified` to make sniffing work with the way I'm routing Tor. In `cuckoo-modified` I had a check to honor the per-VM interfaces. However, when I tried that check upstream, that config option always got applied and overwrote after my code tried to set the interface to be sniffed. Rather than changing more of the scheduling, I figured it was easier/safer to just remove the option. The only reason I kept the KVM global interface setting is to have a fallback, and prevent Cuckoo from scream about a missing config option on startup.

seanthegeek:
@doomedraven This is a port of my PR for the `cuckoo-modified` project. 

https://github.com/brad-accuvant/cuckoo-modified/pull/229

With the way I route Tor, sniffing `virbr0` will only capture broadcast traffic, and traffic to and from the host. Sniffing the VM's NIC directly fixes this, That may be moot once I get VPN routing working (which was not working before I made these changes, and is still not working. I need to do more debugging on that front #1280). This shouldn't break existing routing methods, but more testing would be appreciated.

requirements.txt
seanthegeek: Doing `pip install -r`  with the current requirements from Debain testing fails, and can Break python environments due to incompatibility with the version of OpenSSL in use by the distribution.

modules/machinery/kvm.py
```diff
@@ -3,10 +3,28 @@
...
     dsn = "qemu:///system"
+
+    def _get_interface(self, label):
+        xml = ET.fromstring(self._lookup(label).XMLDesc())
+        elem = xml.find("./devices/interface[@type='network']")
+        if elem is None:
+            return elem
+        elem = elem.find("target")
+        if elem is None:
+            return None
```
seanthegeek: I wrote that function the be a generic NIC finder, for use with Cuckoo and beyond. That case should only occur if someone made a VM without any NICs, in which case they have bigger problems :)

seanthegeek:
No longer needed

jbremer:
@seanthegeek I'd be interested to hear as to why this isn't relevant anymore? :)

PR #1288, opened 2017-02-16
Updates to physical.py to work with FOG 1.3.4

jwilczek:
….3.4.

Added a check to determine the current version of FOG. If the current version is not 1.3.4, a CuckooCriticalError is raised.


jbremer:
Looks good! To be continued as discussed.

docs/book/src/installation/guest_physical/requirements.rst
```diff
@@ -59,3 +59,15 @@ However, any re-imaging platform can be used (Clonezilla, Deepfreeze, etc) to
 accomplish this.
 
 .. _`Fog`: http://www.fogproject.org/
+
+Cuckoo Configuration Requirements
+=================================
+Since we are using physical machines to perform our analysis, we must account
+for the reboot/rebuild time of our physical machines in our Cuckoo configuration.
+Spefically, we must modify the vm_state timeout as specified in conf/cuckoo.conf.
```
jbremer: Typo in `specifically`.

docs/book/src/installation/guest_physical/requirements.rst
```diff
@@ -59,3 +59,15 @@ However, any re-imaging platform can be used (Clonezilla, Deepfreeze, etc) to
...
+	vm_state = 60
+By default, this value is set to 60 (seconds). We need to update it so that it
+reflects the amount of time required to reboot and rebuild the physical guest.
+In testing 10 minutes (i.e., vm_state = 600) has proven sufficient. However,
```
jbremer: Please add something like `with a normal Windows 7 setup and 1 gbit connection`.

docs/book/src/installation/guest_physical/saving.rst
```diff
@@ -48,24 +48,16 @@ image you created above from the "Host Image" option, when finished click the
...
-you wish for the schedule to apply to (``*/5 * * * *``) in the case shown in the
-screenshot below, but you may need to tweak these times for your environment.
-
-    .. image:: ../../_images/screenshots/fog_scheduled_job.png
-        :align: center
-
+Now that you have created and capture an image in FOG, Cuckoo will use this image to
+rebuild the guest machine after each analysis task. If you have provided Cuckoo with 
+valid FOG credentials and enabled Remote RPC (as shown in the Network Configuration
+section), Cuckoo will automatically schedule the Deploy Task in FOG and will also
```
jbremer: Could you leave a link to the `network configuration` section?

modules/machinery/physical.py
```diff
@@ -120,6 +122,12 @@ def stop(self, label):
             # Deploy a clean image through FOG, assuming we're using FOG.
             self.fog_queue_task(label)
 
+            # Hold here until we are certain the physical guest is rebooting
+            while self._status(label) == self.RUNNING:
+                time.sleep(1)
+                continue
+
+
```
jbremer: Please just one newline :)

modules/machinery/physical.py
```diff
@@ -204,13 +214,23 @@ def fog_init(self):
                 "to login into FOG, please configure the correct credentials."
             )
 
+        # Pull out the FOG version from the header and raise error if it is not 1.3.4
+        v = b.find("div",{"id": "version"})
```
jbremer: A space after the comma, please.

modules/machinery/physical.py
```diff
@@ -204,13 +214,23 @@ def fog_init(self):
...
+        runningVer = re.match(r"Running Version\s+(([0-9]+\.)+[0-9]+)",v.text).group(1)
```
jbremer: Same here.

modules/machinery/physical.py
```diff
@@ -204,13 +214,23 @@ def fog_init(self):
...
+        if runningVer != "1.3.4":   # This may be better suited to go in lib.cuckoo.common.constants
```
jbremer: You can remove the comment, I think it's just fine here. We're hardcoding the version number anyway.

modules/machinery/physical.py
```diff
@@ -204,13 +214,23 @@ def fog_init(self):
...
         # for "downloading" a safe image onto the host. Great piece of FOG API
         # usage here.
         for row in b.find_all("table")[0].find_all("tr")[1:]:
-            hostname, macaddr, download, upload, advanced = row.find_all("td")
-            self.fog_machines[hostname.text] = (
-                macaddr.text, next(download.children).attrs["href"][1:],
+            hostinfo, imagename, actions = row.find_all("td")
+
+            self.fog_machines[hostinfo.find("a").text] = (
+                hostinfo.find("small").text, actions.find(title="Deploy").parent.attrs["href"][1:],
```
jbremer: Please put the `actions.find...` on the next line, as it looks like this will be going over our 80-char limit.

nsmfoo:
Not directly related to this PR, but more to the physical setup.  

It seems like the Samba client generates a weird error:
    _"mkdir failed on directory /var/run/samba/msg.lock: Permission denied"_ 
I don't believe that it affects functionality as box gets re-imaged. 

But it's not connected to Cuckoo and can be recreated as follows:

- Error message
$ net rpc info
mkdir failed on directory /var/run/samba/msg.lock: Permission denied
Unable to find a suitable server for domain WORKGROUP

- Try as root, no error message
$ sudo net rpc info
Unable to find a suitable server for domain WORKGROUP

- Now try as your regular user once more.. and no error message =)
$ net rpc info
Unable to find a suitable server for domain WORKGROUP

Sadly this is not reboot safe, but at least the above will remove the error message until that happens, obviously there must be some better way to solve this ;)

nsmfoo:
A suggested (edge-case) stability improvement for this PR.

- Problem:
 If cuckoo crashes and you restart the process. The "auto submission process"(tm) will submit samples to a client regardless of the client state, for example if it's being re-imaged at the time. On VM's this is not a big issue as it's a quick process to restore a snapshot. While with a physical machines it is not. 

- Suggested solution: 
Check that the client is not being re-imaged by FOG before using it

jwilczek:
@nsmfoo I don't think the problem you outlined could ever cause in actual issues. The way Cuckoo is structured, if a job is submitted and subsequently assigned to a machine that is being re-imaged, the job will hold and wait until the machine is available before beginning the job. This has been my experience in my somewhat extensive testing.

Have you encountered something to the contrary?

jwilczek:
@nsmfoo I've encountered the RPC issue - I do not know how to fix that but would be willing to address it if I knew how.

jbremer:
@jwilczek @nsmfoo What do you think about #1092 in addition to this PR?

jwilczek:
First, I've been using the physical setup for a long time and have never seen a machine not reboot. Is it possible there's an issue somewhere with the older version of Cuckoo being used here? 

To me, one of the reasons for a physical setup was to make it look more realistic and harder for code to determine it's in a sandbox. If a procedure is registered, does that give them one more thing to look for?

nsmfoo:
@jwilczek the RPC error message, I'm pretty sure it's new. I did an installation of 14.04 LTS just to check and the same error is present there as well (I think I would have seen it as my previous Physical installation was using 14.04). My guess is that some "new" patch caused this error(?)
 
Could perhaps also be connected to some Apparmor profile update, as I saw someone suggest. 
But as it seems only cosmetic issue, it might not be worth looking into more?

nsmfoo:
@jwilczek I too have had a successful deployment of physical cuckoo for a long time and I'm pretty sure I did not see this previously, sadly the old environment is no more, hence I can't compare old and new.  

TL;DR

I still suggest that there might be room for a check if the current machine is being re-images before it is being assigned a new sample =)

The major change in my environment is that Cuckoo physical is running on a virtual guest, separated from the FOG instance, a physical machine. 

If I only hook up two physical machines and loads four samples, the second pair of samples (3 and 4) is being send for analyses as soon as cuckoo deems that the client is rebooting.  Given enough delay and fast re-imaging I guess that "raise condition" could be handled or pose not problem

An unverified problem could be that the tcpdump starts and captures traffic from the re-imaging, this could perhaps be a problem if Cuckoo and FOG is installed on the same system (?) not seeing this in my environment... so just a guess

jwilczek:
What's the status of this? I had it fixed and working reliably and am willing to update/revisit if needed. Many/many reasons to move to Fog 1.3.4.

jwilczek:
@jbremer I just confirmed that the code for 1.3.4 should work with new cuckoo... however, i'm seeing a weird issue.

`2018-02-01 15:51:22,277 [cuckoo] CRITICAL: CuckooCriticalError: Error initializing machines: Unknown error occurred trying to obtain the status of physical machine cuckoo2. Please turn it on and check the Cuckoo Agent.`

The machine is on and listening on port 8000 (verified i can connect with nc).

Any thoughts?

jwilczek:
Negative, this entire thread is about FOG and using baremetal. The issue ended up being as referenced in #2050, the latest agent does not play well with RPC.  If I go back to the agent of RC2, then all works as expected.

jbremer:
Just one year later, thanks again! Manually merged it.

PR #1293, opened 2017-02-20
Fix Mattermost URL reporting

razuz:
* fixes URL reporting crash
* added new key hash-url in case URL should be hashed
* reports error to log when status code is not 200

Note that POST max size by default is 4000 chars

Cheers,
razu


jbremer:
Thanks!

PR #1307, opened 2017-03-01
Scoring improvements

FernandoDoming:


jbremer:
Closing PR. Not really what I had in mind and if we'd go for something like this (in the near future), then the code could be simplified quite a bit. It's also not entirely complete (missing packages etc).

PR #1319, opened 2017-03-06
Processing module for ExfilDetect.com added.

0x71:
Hi,

I just added a new processing module allowing automated pcap scans for advanced data exfiltration using a new platform called [ExfilDetect](https://exfildetect.com). ExfilDetect is a free online service that analyzes pcap files allowing the identification of hidden data exfiltration.
 
To be more precise, it detects advanced steganographic data exfiltration and protocol tunneling hidden in DNS and ICMP (more protocols are to be added in the future). The automated detection is based on a novel approach investigated during my Masters thesis that will be published in March 2017. The new approach uses behavior-based statistical analysis and machine learning. 

If you are interested in how an analysis looks like, take a look [here](https://exfildetect.com/index.php?page=analyses&query=326932326b39b0bd8188c4fade9ca12b20c6f81075613dfbd7a1bfd900cb42f6).

Best regards,
Nils

### DIFF
--- a/conf/processing.conf
+++ b/conf/processing.conf
@@ -36,7 +36,7 @@ enabled = yes
 enabled = yes
 
 [exfildetect]
-enabled = yes
+enabled = no
 url = https://exfildetect.com
 retries = 5
 # public cuckoo api key
--- a/conf/processing.conf
+++ b/conf/processing.conf
@@ -35,6 +35,13 @@ enabled = yes
 [dumptls]
 enabled = yes
 
+[exfildetect]
+enabled = yes
+url = https://exfildetect.com
+retries = 5
+# public cuckoo api key
+apikey = 4c7318a3c8c6b33b3a0ff6c77aca0423a239cc3603661ec038f7fae1f2a349310c36111919a904767e4758121a0348b4b04996abc64352f83eb3f76d8d316750
+
 [googleplay]
 enabled = no
 android_id =
--- a/modules/processing/exfildetect.py
+++ b/modules/processing/exfildetect.py
@@ -1,0 +1,94 @@
+# Copyright (C) 2010-2013 Claudio Guarnieri.
+# Copyright (C) 2014-2016 Cuckoo Foundation.
+# Copyright (C) 2016-2017 Nils Rogmann
+# This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
+# See the file 'docs/LICENSE' for copying permission.
+
+import logging
+import os
+import requests
+import time
+import json
+
+from lib.cuckoo.common.abstracts import Processing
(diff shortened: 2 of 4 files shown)


conf/processing.conf
```diff
@@ -35,6 +35,13 @@ enabled = yes
 [dumptls]
 enabled = yes
 
+[exfildetect]
+enabled = yes
```
doomedraven: set no by default ;)

0x71:
@doomedraven Thanks ;-)

jbremer:
Although your project looks interesting, we can't integrate it. Currently we don't have integration with any 3rd party services enabled by default (there's just VirusTotal and I disabled it by default just now). This is due to the sensitivity of a lot of our users their Cuckoo instances. I welcome you to implement your ex-filtration analysis ideas in our own `.pcap` reading processing module and we can work from there. See also https://github.com/cuckoosandbox/cuckoo/blob/package/cuckoo/processing/network.py

0x71:
@jbremer Thanks for your feedback! I am aware of the sensitive data that might be generated by some private Cuckoo instances. That is why the source code of my project (the pcap analysis and machine learning part) is to be published under GNU GPLv3. Like for your MISP integration, the users could use this code to run a local analysis instance of ExfilDetect by changing the specified address in conf/processing.conf. Is that a possibility for you? Otherwise, I will consider the integration in your `.pcap` reading module (which is probably not that simple).

jbremer:
Is your library doing any performance / resource needy stuff? E.g., MISP requires a MySQL server, its own API endpoint, etc. If your code could be used in Cuckoo simply by making a Python package out of it and importing that (such as we do with, e.g., my https://github.com/jbremer/httpreplay library), that'd be a much, much easier to setup, manage & maintain add-on as compared to an external service.

0x71:
A Python package sounds good. The framework needs some performance to create the initial baseline (optional) that is optimized for a local network. However, a general baseline that was created during my Masters thesis could be shipped with the package. Currently, the framework requires a MySQL Server to save the baseline, analysis results, and to visualize the results using a web interface. For the Cuckoo integration, I could build a "light version" that can perform pcap analyses without the MySQL Server. Of course, I could leave out the web interface, too. In consequence, this version could basically run with two local python instances (pcap sniffer/analyzer + classification server).

jbremer:
@0x71 Ok, how do you suggest we move forward? Shall I close this issue and you make a new PR once you've got the whole Python package + classification server etc going on?

0x71:
@jbremer That is a good proposal. I will make a new PR for the whole python package.

jbremer:
Ok, closing this one then. Looking forward to it!

PR #1320, opened 2017-03-06
Improvements to human.py

FernandoDoming:


FernandoDoming:
Added functionality to automatically download files in IE11

cuckoo/data/analyzer/windows/modules/auxiliary/human.py
```diff
@@ -151,14 +188,76 @@ def run(self):
...
+            if self.do_change_windows:
+                change_windows()
+
             if self.do_click_buttons:
                 USER32.EnumWindows(EnumWindowsProc(foreach_window), 0)
 
             KERNEL32.Sleep(1000)
+
+    @staticmethod
+    def handle_32770(frameHandle):
```
rholloway: Overall this works pretty well (I'm running variant of this code on a test system). One issue we hit was this would continue to trigger right + enter even after the "View Downloads" run button in IE11 was clicked to run the file. In some cases it would cycle through and keep clicking other things in that "View Downloads" window, most of which unrelated and undesirable.

I had resolved by adding a flag for "view_downloads_clicked" that I set once doing right right + enter enter and checking before, although this may have other side affects if click needs to happen multiple times. My modifications below.

```
@staticmethod
    def handle_32770(frameHandle):
        """
          Handle IE11's "View Downloads" window
        """
        if not Human.view_downloads_clicked:
            log.debug("Setting %s as foreground window", frameHandle)
            USER32.SetForegroundWindow(frameHandle)
            # Press RIGHT + ENTER to run the file
            log.debug("Waiting 15 seconds, hopefully download finishes in time")
            KERNEL32.Sleep(15000)

            log.debug("Sending RIGHT + ENTER...")

            USER32.keybd_event(VK_RIGHT, 0x4D, KEYEVENTF_EXTENDEDKEY | 0, 0)
            USER32.keybd_event(VK_RIGHT, 0x4D, KEYEVENTF_EXTENDEDKEY | KEYEVENTF_KEYUP, 0)
            USER32.keybd_event(VK_RETURN, 0x1C, KEYEVENTF_EXTENDEDKEY | 0, 0)
            USER32.keybd_event(VK_RETURN, 0x1C, KEYEVENTF_EXTENDEDKEY | KEYEVENTF_KEYUP, 0)

            Human.view_downloads_clicked = True
```

cuckoo/data/analyzer/windows/modules/auxiliary/human.py
```diff
@@ -151,14 +188,76 @@ def run(self):
...
+        USER32.SetForegroundWindow(frameHandle)
+        # Press RIGHT + ENTER to run the file
+        log.debug("Sending RIGHT + ENTER...")
+        USER32.keybd_event(VK_RIGHT, 0x4D, KEYEVENTF_EXTENDEDKEY | 0, 0)
+        USER32.keybd_event(VK_RIGHT, 0x4D, KEYEVENTF_EXTENDEDKEY | KEYEVENTF_KEYUP, 0)
+        USER32.keybd_event(VK_RETURN, 0x1C, KEYEVENTF_EXTENDEDKEY | 0, 0)
+        USER32.keybd_event(VK_RETURN, 0x1C, KEYEVENTF_EXTENDEDKEY | KEYEVENTF_KEYUP, 0)
+
+    @staticmethod
+    def handle_ieframe(frameHandle):
```
rholloway: I think this one worked pretty well as well, but vaguely recall in some cases this one does trigger on other toolbars other than download notification toolbar. Not sure if there is much we can do about it though, since I looked into it a bit and seems in >IE8 when IE changed how these windows are generated things got a lot more generic and hard to find and click buttons...

For example, I think this one triggers on bars that pop up for plugins if those haven't been pre-configured in registry (ie Adobe PDF notification bar for defaults if they haven't been set yet, etc)

PR #1323, opened 2017-03-08
Fixed AttributeError when traversing through VMs.

shahar4499:
**Methods '_get_virtual_machines', '_get_virtual_machine_by_label':**
Since the '_vm.summary.config_' instances don't always have a '_name_' attribute, it sometimes fails when traversing through an ESX host with a lot of VMs.
Therefore I caught the _AttributeError_ exceptions.

Tested and worked.

### DIFF
--- a/modules/machinery/vsphere.py
+++ b/modules/machinery/vsphere.py
@@ -234,20 +234,20 @@ class vSphere(Machinery):
 
         for dc, dcpath in traverseDCFolders(conn, conn.content.rootFolder.childEntity):
             for vm in traverseVMFolders(conn, dc.vmFolder.childEntity):
-                try:
+                if hasattr(vm.summary.config, 'name'):
                     self.VMtoDC[vm.summary.config.name] = dcpath
                     yield vm
-                except AttributeError:
-                    pass
+                else:
+                    continue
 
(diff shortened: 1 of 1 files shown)


jbremer:
Why doesn't it have a `name` field? And in this scenario, won't you be missing out on certain VMs (namely, those without a `name`)?

shahar4499:
The fact that they don't always have a `name` attribute is part of the `pyvmomi` module.
The VMs I'm missing out on are probably VMs that I don't have certain permissions on, since I'm not the admin of the VCenter. The machines which I created will always be visible to me, and so, I will not be missing out on them (should have mentioned it, sorry)

jbremer:
I'm not a big fan of `AttributeError` though, could you please do something like the following instead?

```
if not hasattr(vm.summary.config, "name"):
  continue
```

And last but not least requesting some comments from our ESX experts. Any comments, @jgajek @KillerInstinct?

shahar4499:
Changed it to `hasattr`.

jgajek:
Looks good to me!  Thanks @shahar4499 for reporting this!  Feedback from vCenter users in an actual enterprise environment is always highly appreciated.

jbremer:
Thanks! Merging.

PR #1337, opened 2017-03-14
Update Login, Task Table parser and reboot machine request functions to Fog ver 1.3.4

jabex:


jabex:
The main idea is to use features of Fog 1.3.4 to handle the reboot of physical machine. In particular, the new version of the Fog agent install a windows service that handle many operations into the machine. So, when a new Task (deploy or capture) is scheduled, the Fog service handle the reboot request. In this way, the net rpc shutdown request becames the failover operation in case of the Fog Service fault.

jbremer:
Anyone have comments on this PR? And also, do we want to have a FOG service running on the analysis machine?

jabex:
The cuckoo agent may become unuseful, for many reasons, to reboot the machine using the net rpc shutdown request. The FOG service have more resilient than cuckoo agent, and avoid a manual reboot of the analysis machine when the cuckoo agent fails and the FOG deploy has been scheduled.  From my tests (e.g. the analysis of a Ceber sample) I suppose that the main reason is that cuckoo agent is a windows process instead a windows service.

jbremer:
Thanks for your contribution! I'm closing this as duplicate of #1288 (which has now been merged). If there's anything in this PR that you'd like to submit on top of #1288, then please create a new PR.

PR #1349, opened 2017-03-21
Fix rundll32 WoW Redirection

fdivrp:
Currently not able to execute 64bit Dlls on 64bit Windows with 32bit Python.  

This checks if the Dll is 64bit and the machine is 64 bit and Python is 32bit and then uses the Sysnative path to avoid redirection and actually use C:\Windows\System32\rundll32.exe instead of being redirected.

### DIFF
--- a/analyzer/windows/modules/packages/dll.py
+++ b/analyzer/windows/modules/packages/dll.py
@@ -6,6 +6,8 @@
 import os
 import shlex
 import shutil
+import platform
+import pefile
 
 from lib.common.abstracts import Package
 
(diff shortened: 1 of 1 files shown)


jbremer:
Are you sure that `rundll32.exe` doesn't simply invoke its 64-bit version through a subprocess?
I'll have to double check this behavior :-)

fdivrp:
Yes, I believe subprocess is redirected by SysWow64 and you need to use the Sysnative directory in order to call the 64bit rundll32.exe from 32bit Python.

That makes sense to not have that dependancy.  I removed the pefile dependency and updated to manually parse.

PR #1354, opened 2017-03-23
Fix rooter, add memory profile per vm, and linux analyzer related fixes

doomedraven:


cuckoo/apps/rooter.py
```diff
@@ -348,48 +348,52 @@ def cuckoo_rooter(socket_path, group, ifconfig, service, iptables, ip):
...
-                break
-        else:
-            log.debug(
-                "Processing command: %s %s %s", command,
-                " ".join(args),
-                " ".join("%s=%s" % (k, v) for k, v in kwargs.items())
-            )
-
-            output = e = None
+            command, addr = server.recvfrom(4096)
```
jbremer: Please un-indent this entire code block to restore its initial indentation ;-)
Or actually, I'm not sure yet about the whole `EINTR` thing yet.

cuckoo/data-private/cwd/conf/esx.conf
```diff
@@ -56,4 +56,12 @@ resultserver_port = {{ esx[machine].resultserver_port }}
...
 # specific VMs. You can run samples on VMs with tag you require.
 tags = {{ esx[machine].tags }}
+
+# (Optional) Specify a memory profile to be used by volatility for this
+# virtual machine. This will override the guest_profile variable in
+# memory.conf which solves the problem of having multiple types of VM's
+# and properly determining which profile to use. Examples below:
+#mem_profile = WinXPSP2x86
+#mem_profile = Win7SP0x86
+#mem_profile = Win7SP1x64
```
jbremer: One of these should certainly be defined, i.e., `mem_profile = {{ esx[machines].mem_profile }}` etc as otherwise those values are missing from the configuration (which is an error).

cuckoo/processing/memory.py
```diff
@@ -1020,12 +1025,14 @@ def run(self):
...
         self.key = "memory"
+        task_machine = self.task["guest"]["name"]
+        machine_manager = self.task["guest"]["manager"].lower()
 
         results = {}
         if HAVE_VOLATILITY:
             if self.memory_path and os.path.exists(self.memory_path):
                 try:
-                    results = VolatilityManager(self.memory_path).run()
+                    results = VolatilityManager(self.memory_path).run(manager=machine_manager, vm=task_machine)
```
jbremer: I'd propose to do the profile selection in this class already and pass it along properly to `VolatilityManager`, which will do its own thing (namely the whole interaction with Volatility). Don't forget to include some basic unit tests for the `profile` selection process.

cuckoo/processing/memory.py
```diff
@@ -938,14 +938,19 @@ def get_osprofile(self):
...
+    def run(self, manager=None, vm=None):
         results = {}
 
         # Exit if options were not loaded.
         if not self.voptions:
             return
 
-        vol = VolatilityAPI(self.memfile, self.osprofile)
+        # Check if theres a memory profile configured in the machinery config.
+        profile = Config(manager).get(vm).get("mem_profile")
```
jbremer: Use `config("%s:%s:mem_profile" % (manager, vm))` or perhaps some attribute from `self.task` if that's available (that'd be in the `Memory` class below, though), but I'd have to check that myself as well as I'm not sure if it's properly populated at the moment.

PR #1355, opened 2017-03-23
when reprocessing, delete previos report(s), no issues when reports d…

doomedraven:
…oens't exists


jbremer:
Incomplete as you'll also want to drop all related items (probably). See also https://github.com/cuckoosandbox/cuckoo/blob/package/cuckoo/web/analysis/views.py#L359-L421

jbremer:
Still incomplete I think :-) I.e., what should we do with regards to, e.g., linked screenshots etc?
Also I'm wondering if there's such a thing as a reference count in MongoDB/GridFS. If two analyses drop the same binary (a dropped file) and one of these analyses is removed from our database - then the dropped binary will probably not be accessible for the other analysis either.

jbremer:
Closing this for now, will have to be address this some other time.

PR #1368, opened 2017-03-29
processing: clean up temporary file after sorting pcap

chigley:
The files weren't being cleaned up before, which was causing a build-up
of temporary files on our system (one file per analysis).

### DIFF
--- a/cuckoo/processing/network.py
+++ b/cuckoo/processing/network.py
@@ -965,12 +965,12 @@ def batch_sort(input_iterator, output_path, buffer_size=32000, output_class=None
         for chunk in chunks:
             try:
                 chunk.close()
-            except Exception:
-                pass
+            except Exception as e:
+                log.exception("Error closing chunk: %s", e)
             try:
                 os.remove(chunk.name)
-            except Exception:
-                pass
+            except Exception as e:
+                log.exception("Error removing %s: %s", chunk.name, e)
 
 class SortCap(object):
     """SortCap is a wrapper around the packet lib (dpkt) that allows us to sort pcaps
--- a/cuckoo/processing/network.py
+++ b/cuckoo/processing/network.py
@@ -965,6 +965,9 @@ def batch_sort(input_iterator, output_path, buffer_size=32000, output_class=None
         for chunk in chunks:
             try:
                 chunk.close()
+            except Exception:
+                pass
+            try:
                 os.remove(chunk.name)
             except Exception:
                 pass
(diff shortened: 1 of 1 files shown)


jbremer:
Thanks! I'll have a look in a bit.

chigley:
I just updated my branch to improve the commit I'm requesting to be merged.

My previous fix wasn't sufficient because it didn't take into account the fact that multiple temporary files can be assigned. In this case, my previous commit would only delete the last of the temporary files, rather than all of them.

In spending more time looking at this, I realised that the existing code already attempts to handle the file cleanup. However, I found that it wasn't working due to a hidden error in the `chunk.close()` call:

    failed during loop: 'Reader' object has no attribute 'close'

I've updated my commit to make the cleanup more resilient: the temporary files will now always be deleted, regardless of whether `chunk.close()` fails.

Should I make an issue about the hidden "failed during loop: 'Reader' object has no attribute 'close'" error in the `chunk.close()` call?

Thanks for your time!

jbremer:
Hah, good catch! This is a good example of why I don't like `try`/`catch` statements that don't log anything :-)
Do you happen to see the real error? If you'd be so kind to properly fix it that'd be great (and otherwise I can do that as well in a bit) ;-)

chigley:
> This is a good example of why I don't like `try`/`catch` statements that don't log anything :-)

Me neither! I've added a second commit to the PR, which adds some logging.

> Do you happen to see the real error?

Not immediately. I'll update here if I find it :)

jbremer:
@chigley Please take a second look at `dpkt.pcap`'s `Reader` vs `Writer` class. I believe the `Reader` class doesn't have a `close` method - if you could fix the issue properly according to that information, that'd be great :-)

chigley:
I finally found time to look at this. You're right, `fd` can be either a `Writer` or a `Reader`, and only the `Writer` has a `close` method.

I've added a third commit to make sure we only try to call `close` in the `Writer` case. I've also rebased all three on top of the latest master branch, to make sure they'll apply cleanly. The directory structure has changed since my original commits.

Cheers!

chigley:
I can't work out why the CI is failing. I re-pushed the final commit to force a second attempt, but the build failed again.

jbremer:
Yes, that's an unrelated issue that started yesterday when a new version of `codecov` was pushed that's incompatible with our old version of `requests`. So I'm about to bump that.. ;-)

jbremer:
Please find the commit linked above - this PR has been merged manually. It will be featured in the next alpha or stable release. Closing this PR as resolved - thanks again! :)

PR #1369, opened 2017-03-29
Bugfixes to rooter.py and machine.py

HarryR:
This pull request introduces three changes:

 * `rooter.py` - Fixed a bug where lock would not be released on error
 * `rooter.py` - Catch exception if response cannot be sent to client UNIX sock
 * `machine.py` - Allow empty `machines=` in virtualbox.conf before adding first VM

I have brought my https://github.com/HarryR/cockatoo branch up to date with Cuckoo master, these patches are required for basic operation of Cuckoo (Distributed still doesn't work, as will have to submit separate patches for that).

This should be quick & easy to review & merge.

If you have any questions, just ask :)

Cheers - Harry

### DIFF
--- a/lib/cuckoo/core/rooter.py
+++ b/lib/cuckoo/core/rooter.py
@@ -14,7 +14,6 @@ from lib.cuckoo.common.config import Config
 
 cfg = Config()
 log = logging.getLogger(__name__)
-unixpath = tempfile.mktemp()
 lock = threading.Lock()
 
 vpns = {}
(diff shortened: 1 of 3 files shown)


jbremer:
Hey! I remember your PR from a while ago - which I had merged manually, see also 60b633aca622dbc175296435c53691a37390e093.
I'd like to point you to our upcoming Cuckoo `package`, see also #863. As part of this, the `machine.py` stuff has been completely reimplemented and the lock release change has been merged (as per the commit above).
I'm fine with catching the exception for not being able to send a command to the unix socket, but would very much prefer to fix the root cause (as with such an error, Cuckoo may not operate properly). Could you perhaps elaborate on this?

HarryR:
I reset my cuckoo branch to `cuckoo:master`, then re-introduced only the patches which were absolutely necessary for my dockerized cuckoo to work again. 

> see also 60b633a.

This seems to have been un-done in the master branch, so I had to add it back in.

> I'm fine with catching the exception for not being able to send a command to the unix socket, but would very much prefer to fix the root cause (as with such an error, Cuckoo may not operate properly). 

I run `rooter` in a separate container from `cuckoo`, so they don't have a shared `/tmp`. cuckoo connects to rooter via `rooter.sock`, but when rooter in the other container tries to send a reply the return address is a temporary file in `/tmp`. Because this file doesn't exist in the rooter container, but does in the cuckoo container, the reply fails (hence the exception, which includes the socket it tries to respond to, for my reference).

Because of this I introduced `rooter_tmp` which Cuckoo uses as the path for the client end of the connection to rooter.sock - this directory is mounted as a volume between the rooter container & the cuckoo container.

Make sense?

jbremer:
Yes, sorry, that commit was only applied to the package branch.
It does, to some extent. Am I correct to understand that by adding the `rooter_tmp`, the exception handling is now superfluous? I.e., because of the shared `cuckoo_tmp` directory the `.send()` method doesn't throw exceptions anymore.

HarryR:
> I.e., because of the shared cuckoo_tmp directory the .send() method doesn't throw exceptions anymore.

Even if the exception is thrown, it would be nice to know where the client side of the socket is supposed to be, it made it immediately obvious what was wrong, whereas before there was nothing telling me what it was trying to send to.

jbremer:
That I agree with, but the hard error (i.e., Cuckoo crashing) should still take place imo. Now you're logging the error while Cuckoo happily continues, surely making some user out there believe all's fine when it's actually not :-)

jbremer:
Also excellent naming on your repository, `cockatoo` :D

HarryR:
Ok, quick question.

Should I use `package` branch or `master` branch going forward?

I was completely unaware of `package` until now, but my problem is I'm unsure of how to add a pull request onto your pull request and how the workflow would be with that.

HarryR:
Ah, I will move my project over to using cuckoo:package instead of cuckoo:master

This pull request is now invalid

jbremer:
Glad to hear that! Let me know if you run into any problems there.

doomedraven:
this may help

* [README](https://github.com/cuckoosandbox/cuckoo/pull/863) 

```
git clone -b package https://github.com/cuckoosandbox/cuckoo.git cuckoo_package
git clone https://github.com/cuckoosandbox/community.git
cp -rp community/data/monitor cuckoo_package/cuckoo/data/

cuckoo_latest_hash=$(readlink cuckoo_package/cuckoo/data/monitor/latest)
rm -r cuckoo_package/cuckoo/data/monitor/latest
echo $cuckoo_latest_hash > cuckoo_package/cuckoo/data/monitor/latest
cd cuckoo_package
virtualenv cuckoo
. cuckoo/bin/activate

```
`

```
python setup.py sdist
python setup.py develop
python setup.py install
```

* Then uncomment tor/inetsim in rooter.conf, and do the rest of configurations in configs

```
cuckoo rooter
cuckoo web -H 0.0.0.0 -p 8000
cuckoo process auto
cuckoo -d
```

* Help
```
cuckoo --help
cuckoo command --help
```

PR #1400, opened 2017-04-12
Add phrases to human.py

MarkusTeufelberger:
click on "always allow", don't click on "never allow"

In Acrobat Reader, I encountered the following issue:

    [modules.auxiliary.human] INFO: Found button "&Open this file", clicking it
    [modules.auxiliary.human] INFO: Found button "&Open this file", clicking it
    [modules.auxiliary.human] INFO: Found button "&Always allow opening files of this type", clicking it
    [modules.auxiliary.human] INFO: Found button "&Always allow opening files of this type", clicking it
    [modules.auxiliary.human] INFO: Found button "&Never allow opening files of this type", clicking it
    [modules.auxiliary.human] INFO: Found button "&Never allow opening files of this type", clicking it
    [modules.auxiliary.human] INFO: Found button "OK", clicking it

This ended up not opening the file, because it only considered the "open" part. I suspect that a rule to click on "always allow" and not click on "never allow" is universal enough to be useful for everyone.

### DIFF
--- a/cuckoo/data/analyzer/windows/modules/auxiliary/human.py
+++ b/cuckoo/data/analyzer/windows/modules/auxiliary/human.py
@@ -51,6 +51,7 @@ def foreach_child(hwnd, lparam):
         "finish",
         "end",
         "allow access",
+        "always allow",
         "remind me later",
         "save", "sauvegarder"
     ]
@@ -58,7 +59,8 @@ def foreach_child(hwnd, lparam):
     # List of buttons labels to not click.
     dontclick = [
         "don't run",
-        "i do not accept"
+        "i do not accept",
+        "never allow"
     ]
 
     classname = create_unicode_buffer(50)


jbremer:
Just curious, which version of Adobe are you using?

MarkusTeufelberger:
Whatever is installed as recommended in vmcloak, I think 9.0.0.

MarkusTeufelberger:
ping

jbremer:
Oke, thanks. Sorry for the delay. Wouldn't have been able to properly merge this before 3326655b426ccbe6a451e8beb0299b5dd9294c09 anyway, but we can have a look now :-)
Does this question always pop up, or only for PDFs with, e.g., javascript in it? If something specific/else, could you share a hash for testing purposes? Thanks.

MarkusTeufelberger:
This is just an example, the actual change is about these 2 phrases which seem rather universal to me.

FernandoDoming:
Hello @MarkusTeufelberger,

Did you find any real use cases where a program yields a dialog with these phrases? Or is it rather a general addition?

I tried submitting various PDF files (Javascript embedded + vanilla) for analysis to a VM (Win7 x64) with Adobe Acrobat 9.0 (installed both manually and via vmcloak). In all cases the PDF files were fully executed and did not get blocked by any dialog box. In my tests `human.py` did not need any modifications because the aforementioned phrase did not show up. 

If you have any samples or use-cases where this modification is needed please share samples / steps to reproduce.

MarkusTeufelberger:
Yes, see my example above. I even quoted the dialog choices above, googling it leads to https://www.adobe.com/devnet-docs/acrobatetk/tools/AppSec/attachments.html for example. I think it is generic enough to be included by default.

jbremer:
Sure, it is generic enough to include by default, however, as we have to maintain any changes (including this one) for the upcoming years, we'd like to be able to reproduce any and all issues. This allows us to properly identify the problem, inform future users about it (may they run into the same issue, e.g., because of an outdated version of Cuckoo), unit test that the change actually works, and ensure that it remains working in the future when we continue modifying our code base.
Hence @FernandoDoming's (and mine, btw) request for additional information.

MarkusTeufelberger:
Sorry, but I can't share samples. I told you already which program displays this dialog. The underlying issue is that cuckoo will click on anything with the substring "open", which is the wrong behavior in this case. I suspect that you can recreate this by uploading a pdf with a file attachment (not embedded JavaScript, this was @jbremer's suspicion). 

By the way, I find the missing comma after "terminer" a bit suspicious... also "installer" is superfluous (as the "install" substring will always match) as well as a few other words there ("continuer", "end" twice, "scanner"...) and in general only seems to work for english and a few french dialog boxes anyways, so I didn't expect that this file is that well supported (but more of a starting point to write your own phrases - e.g. )

If you feel more confident, I can write a quick unit test to add these dialog choices, I didn't find the place where this file is tested though.

jbremer:
Please find the commit above. We've put in a more appropriate fix that simply configures Adobe to not show this message box and instead go for the `open always` option. Closing PR as resolved. Thanks again!

MarkusTeufelberger:
What about the issues with redundant phrases as well as missing unit tests?

jbremer:
Those would most certainly be welcome. I can add you to our internal build matrix repository where you can PR functional unit tests. As we didn't change anything to `human.py` here it doesn't need additional unit tests. However, any and all future changes to it will require those tests ;-)

MarkusTeufelberger:
The fix seems incomplete as it apparently seems to only white list files formats. What about Powerpoints, Powershell scripts, dll files... and once these are added it still is whack-a-mole. It is a partial solution but doesn't solve the underlying issue (accept pop-up windows correctly).

I'll just maintain my patch in private then, thanks for resolving this PR.

PR #1426, opened 2017-04-18
Some basic template edits to add route information 

lowson:
These changes edit the summary and post page to add more details about the route option selected for historical review. I also added the machine name to the post page, typically needed if you plan to interact with the specific machine.


jbremer:
Thanks! I've manually merged this - removed the machinery blob (that wasn't displayed anymore anyway) and added the `routing` column to the `Information of Execution` blob. Closing PR.

PR #1435, opened 2017-04-19
Inetsim2

doomedraven:


enzok:
I think it's still desirable to keep the inetsim dns port configuration option. There is the use case where if inetsim is running on the vm interface, then port 53 is not available as it's already in use by the VM server. It could then be passed to dns_forward() as an argument.

doomedraven:
I tried it, but then you should extend iptables to prevent a lot of leaks, so better just keep it in vm for example ubuntu server or any other soft dists and forget sbout it :)

seanthegeek:
@doomedraven Looks like you have a syntax error near line 361 of `scheduler.py`.

doomedraven:
em why do you think so? i have it in production as it works just fine

doomedraven:
lol you was right, i should fix it in production and forget add it here somehow %)

seanthegeek:
@doomedraven 

I noticed that you removed the setting for the inetsim DNS port. I know you use KVM. How are you deploying inetsim? I've always hosted it on the Cuckoo host on `virbr0`, but libvirt always starts dnsmasq on the virtual bridge interfaces, so port 53 was always taken. I had to host inetsim's DNS on an alternative port (e.g. 5353) and redirect traffic from port 53 to this port using iptables rules.

Also, `test_rooter.py` needs to be updated
https://travis-ci.org/cuckoosandbox/cuckoo/builds/236838286#L3447

doomedraven:
Let me try to explain the situation, if you host inetsim in cuckoo host, the rooter is not hardened for that, so you need add more iptables rules, is much more easier get ubuntu server(not remnux, or similar, think in saving resources) so to avoid leaks of traffic/similar is much more easier to do that, but if you want i can add it, i also prefer be able configure ip/port, but @jbremer still ignoring me :D

seanthegeek:
I only have a basic understanding of `iptables`. Can you explain how traffic to a host-based inet inetsim could leak? 

In cuckoo-modified I used these commands at boot to allow inetsim to work on the host, and I never observed any leakage

```bash
#!/bin/bash

VIRBR_IP="192.168.100.1"
INETSIM_DNS_PORT="5353"
VIRBR_DEV="virbr0"

# Redirect libvirt VM DNS quires to inetsim's DNS port
/sbin/iptables -t nat -C  PREROUTING -d $VIRBR_IP -p udp --dport 53 -j REDIRECT --to-ports $INETSIM_DNS_PORT >/dev/null 2>&1
if [ $? -ne 0 ]; then
    /sbin/iptables -t nat -I PREROUTING -d $VIRBR_IP -p udp --dport 53 -j REDIRECT --to-ports $INETSIM_DNS_PORT
fi

# Allow inetsim to accept traffic for any IP address
/sbin/iptables -t nat -C PREROUTING -i $VIRBR_DEV -j REDIRECT >/dev/null 2>&1
if [ $? -ne 0 ]; then
    /sbin/iptables -t nat -A PREROUTING -i $VIRBR_DEV -j REDIRECT
fi
```

jbremer:
The time has come, this PR has been merged @doomedraven! Please test the latest `master` branch though, to ensure that everything works as expected.

doomedraven:
Yey i thought that time will never come hahhah, will test next week, feel free to close if nobody complain

jbremer:
The new alpha release `2.0.4a3` includes this PR. Please give it a try @doomedraven @seanthegeek @pope-es @enzok! Closing this PR as such. Please leave any feedback in a new github issue or reopen this PR, thanks!

netpacket:
@jbremer 
I'll take a look at it tomorrow. Thanks for pushing out a patch on this.

netpacket:
Looks like InetSim inside a VM works with Cuckoo 2.0.3 without the Cuckoo daemon running.

However, Cuckoo daemon running causes an argument error while running InetSim routing. I could have had an error while installing 2.0.4a3. I'll try checking that.

doomedraven:
>> argument error while running InetSim routing

log?

netpacket:
Here's the logs:
[cuckoo_log_20170628.txt](https://github.com/cuckoosandbox/cuckoo/files/1109542/cuckoo_log_20170628.txt)
[cuckoo_rooter_20170628.txt](https://github.com/cuckoosandbox/cuckoo/files/1109544/cuckoo_rooter_20170628.txt)

jbremer:
That's certainly not the latest version, no ;) You'll have to upgrade successfully before using inetsim.

netpacket:
In order to upgrade, I need to clone the newest release and build it from source, right? 
pip install -U cuckoo seems to be pulling up cuckoo 2.0.3.

st11cks:
I'm also looking for best way to apply changes. would a simple replacement of the 4 revised files work? if so, what about the .pyc versions? ignore, or rename a duplicate of new files with .pyc extension? but they don't appear to be equal in size, at least not for apps/rooter.py and apps/rooter.pyc

pope-es:
Hi st11cks

Today I learnt... that you can create a new virtual environment, and inside it do: pip install -U 'cuckoo==2.0.4a3'

From there you can test.

I will try myself to submit results. Problem is, I've been tampering with a few things in my cuckoo setup for the last days (hello Moloch!). Will probably revert my cuckoo VM to a previous snapshot and try from there.

Cheers

P!

pope-es:
Well, Cuckoo and Inetsim seem to be getting along well. This code works for me guys, how about the others?

I'm getting OK the results in the "Networking" tab. However, no networking activity is shown at the bottom of the "Summary" tab - don't know if this is normal. Screenshots attached. I also have nothing in the "networking" part of the behavioural analysis.

screenshot1
screenshot2

Cheers

Pope

NOTE: I have a rare setup - my Cuckoo thinks he's communicating with a VMware Workstation, but he is really interacting with a free ESXi via SSH, through a wrapper I wrote. Should not affect these results, though. I'm thinking about patching Cuckoo to support this natively, guess could be interesting for other people?

[analyzer.log.txt](https://github.com/cuckoosandbox/cuckoo/files/1116964/analyzer.log.txt)
[cuckoo.log.txt](https://github.com/cuckoosandbox/cuckoo/files/1116965/cuckoo.log.txt)

pope-es:
BTW don't trust the "June 23rd" timestamps in my attached logs, it's because reverted the VM and the clock didn't update. These tests were actually carried out today.

PR #1443, opened 2017-04-20
Changing s/loose/lose/ where it makes the most sense

Popsiclestick:
Fixing a minor typo


jbremer:
The typo in `vendor.js` is actually from a 3rd party dependency, so changing it here wouldn't help much (as the next build of `vendor.js` will restore it). Could you get rid of that one? Thanks!

Popsiclestick:
Changes made!

Issue #1496, opened 2017-04-27
VM not routing via Tor


FoxyNPF:
Hi I have followed the latest instructions for routing via Tor however I cannot get it to work:

https://cuckoo.sh/docs/installation/host/routing.html#routing-tor

My amended routing.conf file:
![1](https://cloud.githubusercontent.com/assets/26876090/25484441/f848a680-2b51-11e7-8985-a78cf8224568.png)

/etc/tor/torrc updated:
![2](https://cloud.githubusercontent.com/assets/26876090/25484450/0192c9a0-2b52-11e7-9c7d-98e8af81d521.png)

cuckoo rooter running and tor restarted:
![3](https://cloud.githubusercontent.com/assets/26876090/25484453/02f08bb6-2b52-11e7-85e1-8fe31178ccf3.png)

After following the instructions I then go open up my VM and go to check.torproject.org and get the following page:

![4](https://cloud.githubusercontent.com/assets/26876090/25484476/1ae5a332-2b52-11e7-89a4-35c32a8e78c8.png)

Is anybody aware of anything that I am missing?

Thanks

doomedraven:
activate verbose on rooter `cuckoo -d rooter` and post output after analysis, try submit url for example `ipinfo.io`

FoxyNPF:
Hi, here is the output from cuckoo -d rooter:

cuckoo@necurs:~$ sudo cuckoo -d rooter
[sudo] password for cuckoo: 
2017-04-27 14:11:37,119 [cuckoo.apps.rooter] DEBUG: Processing command: forward_drop  
2017-04-27 14:11:37,136 [cuckoo.apps.rooter] DEBUG: Processing command: state_disable  
2017-04-27 14:11:37,161 [cuckoo.apps.rooter] DEBUG: Processing command: state_enable  
2017-04-27 14:11:49,932 [cuckoo.apps.rooter] DEBUG: Processing command: forward_drop  
2017-04-27 14:11:49,947 [cuckoo.apps.rooter] DEBUG: Processing command: state_disable  
2017-04-27 14:11:49,971 [cuckoo.apps.rooter] DEBUG: Processing command: state_enable  
2017-04-27 14:11:52,036 [cuckoo.apps.rooter] DEBUG: Processing command: forward_drop  
2017-04-27 14:11:52,051 [cuckoo.apps.rooter] DEBUG: Processing command: state_disable  
[...]

Issue #1440, opened 2017-04-20
inetsim routing option errors


seanthegeek:
When selecting the inetsim routing option, cuckoo geives the following errors, and the task never completes.
```

2017-04-20 10:28:26,883 [cuckoo.core.rooter] WARNING: Rooter returned error: remote_dns_forward() takes exactly 4 arguments (3 given)
2017-04-20 10:28:26,898 [cuckoo.core.guest] INFO: Starting analysis on guest (id=win7, ip=192.168.100.42)
2017-04-20 10:31:29,586 [cuckoo.core.scheduler] ERROR: Error from the Cuckoo Guest: The analysis hit the critical timeout, terminating.
2017-04-20 10:31:30,010 [cuckoo.core.scheduler] ERROR: Failure in AnalysisManager.run
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/cuckoo/core/scheduler.py", line 697, in run
    self.launch_analysis()
  File "/usr/local/lib/python2.7/dist-packages/cuckoo/core/scheduler.py", line 621, in launch_analysis
    self.unroute_network()
  File "/usr/local/lib/python2.7/dist-packages/cuckoo/core/scheduler.py", line 360, in unroute_network
    str(cfg.resultserver.port))
AttributeError: 'Config' object has no attribute 'resultserver'
```

doomedraven:
https://github.com/cuckoosandbox/cuckoo/issues/1380

jbremer:
Yeah, inetsim isn't working yet, but @doomedraven is working on it as you can see :)

doomedraven:
test this PR and feedback is welcome https://github.com/cuckoosandbox/cuckoo/pull/1435

seanthegeek:
@doomedraven @jbremer Hmm, now this is odd.

My usual hack for getting inetsim working in legacy Cuckoo isn't working Cuckoo 2.0.1.

I have a script that runs the following commands at boot:

```bash
VIRBR_IP="192.168.100.1"
INETSIM_DNS_PORT="5342"
VIRBR_DEV="virbr0"

# Redirect libvirt VM DNS quires to inetsim's DNS port
/sbin/iptables -t nat -C  PREROUTING -d $VIRBR_IP -p udp --dport 53 -j REDIRECT --to-ports $INETSIM_DNS_PORT >/dev/null 2>&1
if [ $? -ne 0 ]; then
[...]

Issue #1380, opened 2017-04-07
Cuckoo Package: INetSim REMnux remote_dns_forward() error


Darjuw:
I am trying to get Cuckoo to work with REMnux InetSim on the package version. But for some reason, I'm constantly getting remote_dns_forward() error.

My setup
```
**Host** 172.16.76.1 (Ubuntu 14.04)
**Malware Sandbox VMware** 172.16.76.90 (vmnet1)
**REMnux InetSim VMware** 172.16.76.128 (Have tried NAT, vmnet1, host-only)
```

My configuration
```
[inetsim]
# Route a VM to your local InetSim setup (could in theory also be any other
# type of web service / etc).
enabled = yes
server = 172.16.76.128
```

/etc/inetsim/inetsim.conf (on 172.16.76.128)
```
start service for everything
default service bind and dns address (I have tried setting it to 172.16.76.128 and 172.16.76.1 too but getting same error)
```

/etc/inetsim/inetsim.conf
`default configuration (I run INetSim using sudo inetsim)`

**Errors**

admin2@admin2-7010:~$ sudo cuckoo rooter -g admin2
```
2017-04-07 11:02:48,205 [cuckoo.apps.rooter] ERROR: Error executing command: remote_dns_forward() takes exactly 4 arguments (3 given)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/Cuckoo-2.0.0-py2.7.egg/cuckoo/apps/rooter.py", line 388, in cuckoo_rooter
    output = handlers[command](*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/Cuckoo-2.0.0-py2.7.egg/cuckoo/apps/rooter.py", line 202, in inetsim_enable
    remote_dns_forward(ipaddr, inetsim_ip, "-A")
TypeError: remote_dns_forward() takes exactly 4 arguments (3 given)
```

admin2@admin2-7010:~$ sudo cuckoo 
```
Cuckoo Sandbox 2.0.0
 www.cuckoosandbox.org
 Copyright (c) 2010-2017

 Checking for updates...
 You're good to go!
2017-04-07 11:00:38,947 [cuckoo.core.startup] INFO: Updated running task ID 59 status to failed_analysis
2017-04-07 11:00:39,080 [cuckoo.core.scheduler] INFO: Using "vmware" as machine manager
[...]

PR #1510, opened 2017-04-30
Fix Cuckoo Rooter (Internet, TOR, inetsim) #1440 #1380 #1496

super-a1ice:
If you do want to use the Internet then you need open port 53.

- Internet
    - ACCEPT udp 53
- InetSim
    - ACCEPT udp 53
    - FORWARD port 53
- Tor
    - ACCEPT udp 53

### DIFF
--- a/cuckoo/apps/rooter.py
+++ b/cuckoo/apps/rooter.py
@@ -199,7 +199,13 @@ def inetsim_enable(ipaddr, inetsim_ip, resultserver_port):
     run(s.iptables, "-A", "OUTPUT", "-m", "state", "--state",
         "INVALID", "-j", "DROP")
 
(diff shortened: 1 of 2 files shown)


doomedraven:
See inetsim2 pr

jbremer:
@tatsui @doomedraven does this PR provide additional functionality on top of #1435?

doomedraven:
@tatsui can respond about internet enable and disable, as with host only i think that useless, and there only aprt of that are fixes for inetsim :p, but lets see what tatsui mean with that funcs

jbremer:
Thanks! Closing PR as it doesn't seem to provide additional functionality on top of #1435. Please let us know if that's not the case, though.

PR #1519, opened 2017-05-03
Update _irma.html

r00t0vi4:
When irma checks file with antiviruses and marks file as malisious we see table with results. 
If file is clean we see result "No IRMA results available.". But if irma was terminated we see "No IRMA results available." too.
I delete string "and report.analysis.irma.status" to display table with antiviruses if probe results is clean.


jbremer:
Sounds good to me, thanks! Sorry for delay in merging.

jbremer:
Merged manually and will be available in the next alpha or stable release.

PR #1521, opened 2017-05-04
Update irma.py

r00t0vi4:
Added timeout checks for a while loop.
Without this check when irma server breaks, the analyse goes into an infinite loop.

### DIFF
--- a/cuckoo/processing/irma.py
+++ b/cuckoo/processing/irma.py
@@ -67,7 +67,7 @@ class Irma(Processing):
 
         result = None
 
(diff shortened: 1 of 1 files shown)


jbremer:
Interesting - makes sense. Will check it out later, thanks!

jbremer:
Also fine by me, merging. Thanks!

jbremer:
Actually, this change will also force Cuckoo to *always* wait `self.timeout` seconds - even if the IRMA results are available earlier ;-) Let me fix that.

jbremer:
Merged manually and will be available in the next alpha or stable release.

PR #1564, opened 2017-05-16
Change iptables-persistant to persistent

NaykiSec:
Error in the package name the good one is iptables-persistent
(checked with apt-cache search iptables-persi*)


jbremer:
Thanks! :-) Let us know if you have additional feedback.

PR #1576, opened 2017-05-18
Fixed misleading typo

ThrosturX:
Changed "peristant" to "persistent" in "iptables-persistent" so that copy-pasters will not cry.


jbremer:
Heh, see also #1564. Thanks!

PR #1583, opened 2017-05-19
udev interface-naming change comment |ens33 v eth0

craigdods:


jbremer:
I've seen plenty of other interface names other than `ens33`, so please update the message to be a bit more generic :-)

craigdods:
@jbremer - ens33 is the default naming scheme for debian and ubuntu. I haven't checked other distros right now but I'm relatively certain udev + kernel 4.10+ is going to give you ens33 as your first interface (aka what eth0 used to be).

doomedraven:
Nop, the ens part is default, bit the mumber is different

jbremer:
And also I've seen `em1`, `em2`, etc interface names. Or `enp0s3` if you're running in VirtualBox (not to suggest that running Cuckoo in vbox is a good idea, though, just that there are some variations possible).

erhan-:
enp0s25 on arch and I had different interface names before.

jbremer:
I've slightly updated this PR but included it nonetheless. Thanks again and sorry for the delay in merging! I'll try to synchronize it with our official documentation soon'ish.

PR #1585, opened 2017-05-20
Add basic machinery for OpenStack Nova

robputt796:
Hi,

This adds a basic machinery for interacting with OpenStack clouds rather than traditional virtualisation, in it's current implementation it treats the OpenStack cloud simply like a normal hypervisor maintaining a list of preconfigured virtual machines (server instances) and once the analysis is complete it simply reimages the machine back to the known state.

However long term I'd like to expand this so that:
- Machines autoscale based on the demand of cuckoo, the more analysis tasks you create the more machines it will spawn.
- Machines have a seggregated network using tenant networks in Neutron (currently they all sit on one tenant network)
- Support for Ironic so you can run machinery against bare metal

Currently works and tested against Rackspace Public Cloud (with auth mod switch in config) and Datacentred OpenStack Public Cloud (Vanilla APIs). However of course actual malware analysis on a public cloud may be against the providers Terms of Service (I only used benign files during testing) and the use case would probably be against an OpenStack Private Cloud in reality.

Please let me know if you have any questions, more than happy to submit ammendments. More PRs to follow but I wanted to get the basics in as a starting point.

Best Regards,

Rob


robputt796:
Hi,

I am not sure how to go about satisfying these tests, this config framework is not familiar to me, please can someone give some pointers?

Best Regards,

Rob

jbremer:
Cool! Looping in @geudrik for feedback :-)
Some unit tests are failing due to missing configuration migration rules & additional unit tests for that, but we can resolve that for you.

geudrik:
@robputt796 why not use any of the available OS libs to handle API requests? `shade` is pretty common, as is the native `python-openstackclient`. That way you can get your authorization code out of cuckoo and make future API changes easier to live through.

robputt796:
Thanks for the reply @geudrik.

Personally I am not a massive fan of the Python OpenStack client or indeed the older Python Nova Client and so on, they seem great for CLI but actually using them within Python has never really gone to plan for me. However I will consider looking at shade, I have never heard of it before but reading the docs it seems reasonable.

Best Regards,

Rob

robputt796:
Just to check, is the preference to refactor this into shade before merge? Or would you just like to see this in some iteration? I expect this will be an ongoing thing as I'd like to make a mechanism for autoscaling rather than having to define all machines individually.

geudrik:
@robputt796 Jurriaan tagged me in this as he and I had been emailing about my interest in building out OS support. Autoscaling is doable, but not without significant additions/changes to the underlying codebase (which when I was adding in my own OS support, was something I was trying to avoid). I'll let @jbremer cover interest for scaling support.

Using a lib over doing your own API calls ensures that your code can grow to meet changing API versions without having to be rewritten. I also had chose to mirror the VBox configuration as close as possible (for consistency), so instead of relying on UUIDs for machine management, you could rely on the machine name (much more natural). This also is what you'll want to rely on if you start go down the rabbit hole of scaling horizontally (IMO)

Alpha POC Code sample I was playing with below. Don't use this as an end-all by any means, but this was the direction I was heading. Ping me directly if you'd like to work on this in tandem 😄 
```python

    def _make_nova_session(self, thread_id=None):

        self._nova_sessions = {}
        tid = thread_id or threading.current_thread().ident

        try:
            client = novaClient(
                2,
                self.options.openstack.username,
                self.options.openstack.password,
                self.options.openstack.project_id,
                self.options.openstack.auth_url,
                insecure=insecure
            )
        except Exception as e:
            raise CuckooMachineError(“”"Could not create a Nova session, {}. Ensure that all configuration parameters
                are correct, and that the OpenStack hostname is resolvable (dns must work correctly)“”".format(e))

        self._sessions[tid] = client
        return client

    @property
    def session(self):
        tid = threading.current_thread().ident
        session = self._sessions.get(tid, None)
        if session is None:
            session = self._make_nova_session()
        return session

    def _get_vm(self, vm_ip):
        “”" vm_ip The IP of this VM
        “”"
        ret = self.session.servers.list(search_opts={‘ip’:vm_ip})
        if len(ret) > 1: # Should never happen
            raise CuckooMachineError(“More than 1 VM was returned for ip {} Herm....“.format(vm_ip))
        elif len(ren) < 1: # No IP matching the supplied IP
            raise CuckooMachineError(“No VMs matched ip {}“.format(vm_ip))
        return ret[0]

    def _get_machine_by_label(label):
        “”" Get a Machine() instance by machine label
            cuckoo/core/database.py:82
        “”"
        machine = self.db.view_machine_by_label(label)
        return machine
```

geudrik:
I need to make it through the workweek before I can jump back into this, but I'll spend some time using `shade` and porting some of this code over. See how my stab holds up 😄 

Still unresolved is memory dumping, which is big. There are several ways to tackle this, but for phase one, I'm just going to focus on getting basic functionality going. Perhaps one day I'll learn how to do some kernel programming and help make @jbremer make the on-host module ❤️

jbremer:
@robputt796 @geudrik did we have any progress on this matter? Or what's the latest consensus on approach etc?

geudrik:
@jbremer I've been tied up with other things and @rholloway has been as well. This is still very much on our radar though. Not sure what @robputt796 has plans for, but if he's interested in working with me, I'd love to tag team it!

Memory dumping can wait for phase 2 (IMO) provided that we raise a NotImplemented() and make sure we have a note somewhere (maybe the machinery config) that mentions it's not implemented.

rholloway:
@robputt796 how is your setup going with openstack? Have you done anymore work on it or is this PR current? @geudrik and I also have some openstack machinery we have set up and been running live as beta for analysis, but it's done a bit differently (ran about 1000 samples through it past few days)

* machinery puts a heavier focus on cinder / volumes rather than nova. That is, we base new instances off a cinder volume snapshot rather than a nova snapshot. This was primarily to suit our use case a bit better with a ceph storage backend and our version of openstack which didn't support copy-on-write for nova snapshots (anything prior to Mitaka suffers from this, I believe)

* we use a separate neutron analysis network as you mentioned, worked through the kinks with tcpdump / routing as well (analysis network basically routes through cuckoo host as default gw for tcpdump)

* We spin up instances dynamically rather than statically when a new job comes in, so we don't have to have those VM uuid pre-configured with snapshots in advance. Instead, we just launch a new instance based on the cinder volume snapshot. The only "static" aspect is that we do need to add a section in configuration for each VM we want running simultaneously with label + ip + snapshot id, but the machine / ip gets allocated automatically at least. It would be nice to adjust more towards a profile model where we just have sections for profiles and a max number of VMs to use, but I think that's out of scope currently (although better profile support may be on roadmap for cuckoo somewhere down the line). However, it is still pretty low maintenance and easy to add 100 more potential VMs quickly if desired.

This doesn't quite solve the auto scaling per say, but it sort of does. The configuration basically sets the max VMs that can run at a time, but the VMs aren't running when idle. The downside is any openstack deploy that may not be using copy-on-write storage underneath may suffer since launching an instance may take longer.

robputt796:
@rholloway Your machinery sounds like what I was hoping todo, are you likely to release it to the community so we can all drool over it?

jbremer:
@robputt796 @rholloway @geudrik any update here?

robputt796:
Yeah no real updates, the good news is I do have a new OpenStack Pike mini cloud setup in my house so I at-least have a decent place to test this. I probably need to come back to this. Really what we need is to see @rholloway's stuff :-)

rholloway:
Same here, I need to get back to it. It's running off and on, but due to some other issues and priorities I keep getting pulled away from validating the last bits. We did run into a couple hiccups that may or may not be fixed with old volumes not being deleted promptly/quickly in cinder due to some of our configurations/mysql driver for cinder (unrelated to cuckoo openstack, but indirectly affected it's use for a while).

I'll try to revisit and open a PR for our version soon and we can compare notes. I think the `use_cinder` and `use_glance` sounds like a great idea.

PR #1586, opened 2017-05-20
Add mongodb authentication if username and password is set.

neveralso:
Add mongodb authentication if username and password is set when executing `cuckoo clean`.


jbremer:
Please find & evaluate the following commit https://github.com/jbremer/cuckoo/commit/0314f11336410eccc400d3e04e347ae108ceed1d. As per the commit message it should be a replacement for this PR. Closing as such - sorry for the delay. Do let us know if you run into any other issues.

PR #1587, opened 2017-05-22
Added external customization guide

Tigzy:
Added customization guide


jbremer:
If you're willing to rewrite your blogpost to something that can be embedded in our documentation *directly* I'll be happy to take it. Or if you'd like to do a guest blog post on our near-future website, that's certainly also possible (reach out to me for that). However, as-is we don't have any 3rd party links in our documentation, so for now I don't think that's the way to go.
Let us know what it's going to be, but I'll be closing this PR for now.

PR #1614, opened 2017-05-27
opera already uses webkit

doomedraven:
https://www.siliconrepublic.com/life/opera-hits-300m-users-switches-to-webkit-and-chromium

WebKit is a layout engine software component for rendering web pages in web browsers. It powers Apple's Safari web browser. It was forked by Google to create Blink, used by Chromium-based browsers such as Google Chrome and Opera.
source: https://en.wikipedia.org/wiki/WebKit

### DIFF
--- a/cuckoo/web/src/scripts/app.js
+++ b/cuckoo/web/src/scripts/app.js
@@ -124,7 +124,7 @@ class CuckooWeb {
     // recommended browser list.
     static isRecommendedBrowser() {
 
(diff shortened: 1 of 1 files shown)


jbremer:
Thanks! Will merge it after finishing off some other stuff.

jbremer:
Manually merged it now, thanks again. Also included the reference in `cuckoo/web/static/` ;-)

Issue #1280, opened 2017-02-09
Routing to VPN not working


seanthegeek:
I'm trying to get VPN routing working. I installed OpenVPN and resolvconf, configured and started an OpenVPN client connection at `/etc/openvpn/us.conf`:

```client

dev tun0

remote [REDACTED] 1300
cipher AES-256-CBC

# Tunnel IPv6 traffic as well as IPv4
tun-ipv6

# Ignore routes
route-nopull


# Keep trying indefinitely to resolve the
# host name of the OpenVPN server.  Very useful
# on machines which are not permanently connected
# to the internet such as laptops.
resolv-retry infinite

# Most clients don't need to bind to
# a specific local port number.
nobind

# Try to preserve some state across restarts.
persist-key
persist-tun

# Enable compression on the VPN link.
comp-lzo

# Set log file verbosity.
verb 3

remote-cert-tls server

ping-restart 60

# Allow calling of built-in executables and user-defined scripts.
script-security 2

# Parses DHCP options from openvpn to update resolv.conf
up /etc/openvpn/update-resolv-conf
down /etc/openvpn/update-resolv-conf

ping 10

[REDACTED KEYS]
```

Configured Cuckoo's `kvm.conf`:

'''[vpn]
# By default we disable VPN support as it requires running utils/rooter.py as
# root next to cuckoo.py (which should run as regular user).
enabled = yes

# Comma-separated list of the available VPNs.
vpns = us

[us]
# Name of this VPN. The name is represented by the filepath to the
# configuration file, e.g., cuckoo would represent /etc/openvpn/cuckoo.conf
# Note that you can't assign the names "none" and "internet" as those would
# conflict with the routing section in cuckoo.conf.
name = us

# The description of this VPN which will be displayed in the web interface.
# Can be used to for example describe the country where this VPN ends up.
description = United States

# The tun device hardcoded for this VPN. Each VPN *must* be configured to use
# a hardcoded/persistent tun device by explicitly adding the line "dev tunX"
# to its configuration (e.g., /etc/openvpn/vpn1.conf) where X in tunX is a
# unique number between 0 and your lucky number of choice.
interface = tun0

# Routing table name/id for this VPN. If table name is used it *must* be
# added to /etc/iproute2/rt_tables as "<id> <name>" line (e.g., "201 tun0"). 
# ID and name must be unique across the system (refer /etc/iproute2/rt_tables
# for existing names and IDs).
rt_table = tun0'''

[...]

PR #1621, opened 2017-05-31
Make rooter insert rules instead of adding them

seanthegeek:
By default, libvirt applies the following iptables rules to host-only
bridges on startup:

```
-P FORWARD DROP
-A FORWARD -i virbr0 -o virbr0 -j ACCEPT
-A FORWARD -o virbr0 -j REJECT --reject-with icmp-port-unreachable
-A FORWARD -i virbr0 -j REJECT --reject-with icmp-port-unreachable
```

Rooter must use the insert (-I) action rather than append (-A) so that the
new rules have a higher priority than existing ones. Otherwise,
forwarding and other commands will not work.

This is the root cause and fix for the routing problem I was having in
issue #1280.


jbremer:
Good stuff, thanks again! Manually merged it and will be part of the next alpha or stable release.

jbremer:
On second thoughts, no, this PR completely breaks our per-analysis network routing. Most of the rules are very dependant on the exact order in which they are. By using `-I` rather than `-A` you'll effectively reverse that order. In order to properly perform this PR, which is something we can work towards, we'll have to modify it a little bit.
Using rule numbers, i.e., `-I <table> <num>`, doesn't seem to be a good way - provided that other Cuckoo analyses and/or tools may be altering the `iptables` rules as well.
Most likely that leaves us with reversing the insert operations for our `iptables` entries. Naturally this should be tested well, but I'll help there.. ;-)

seanthegeek:
@jbremer At analysis time the iptables forwarding commands are only for that VM's IP address, so there shouldn't be race conditions or broken routing wwith this PR as-is.

seanthegeek:
Closing, as this is cor longer needed as of 2.0.3a4. VPN works. However, there are still some issues with inetsim when ran on the local system. I'll create a speperate issue for that.

doomedraven:
yep as i told i totally not recommend run it in the same host to avoid leaks as we spoke in PM :)

PR #1623, opened 2017-06-01
reorder host installation doc pages

daanfs:
It makes sense to put the "Configuration" and "Cuckoo Working Directory" pages before the "Per-Analysis Network Routing" page.

The current order seems a little bit confusing on a read through, since the routing page mentions a few configuration directives before even mentioning the config files.


jbremer:
Makes sense, yes, thanks! Merged and all that.

PR #1625, opened 2017-06-01
Begin adding in Yara rules

kevross33:
First one is to detect ransomware messages in memory. Sometimes you find this even if the ransomware doesn't run right.


jbremer:
Looks fine to me, however, would you be so kind to PR this against `community`? :)

jbremer:
Bump @kevross33.

PR #1657, opened 2017-06-11
Also show HTTP(S) when only HTTPS is present

mvdnes:
When there is only HTTPS, but not HTTP traffic, the total at the top of the page would incorrectly show a count of 0.


jbremer:
Good catch, thanks! Will merge soon'ish.

jbremer:
All merged up now, thanks again!

PR #1670, opened 2017-06-21
Close office after some time to trigger document_close hooks

FernandoDoming:


jbremer:
This has been included in our `2.0.4a3` release, works well :-)

Issue #1348, opened 2017-03-21
Status of vCenter machines is not reliable


shahar4499:
**The Issue:**
I'm using the VSphere machinery module.
When consuming the REST API to check a machine's status, I sometimes get the previous status, or just a status that is inaccurate.
For example I got "poweredOn" when the machine was off for quite a while, and I got "null" or "suspended" instead of "poweredOff".

**When it can happen:**
I don't have an exact way to reproduce this, but here are some examples for when I got it:
- When cuckoo encounters an unexpected error
- When interrupting cuckoo before it finishes a task

**Reflections**
As I see in the code, the Flask API only queries the DB, and it doesn't make a new request to VSphere API in order to get the current status.
This may save a tiny bit of load, but it is highly unreliable.
I personally think cuckoo should query the VSphere API each time.
I'm still not familiar enough with the code to fix this completely by myself, but with a little bit of guidance, I'd love to.

jbremer:
Perhaps I should start with: what's your goal and why is it important to know the exact machine status?

shahar4499:
I'll try to spare you the details. I'm integrating Cuckoo to a large system. The priority queue of the analyses is managed by our own server, and it sends tasks to Cuckoo with the API, after querying it for an available machine. If the status of the machine is powered off, it won't be used (and usually it gets stuck with this status until restarting Cuckoo).
We have some good reasons that our server shouldn't query the VSphere API by itself (both design-wise and security-wise).

jgajek:
@shahar4499 Thanks for the report -- I'll take a look.

jbremer:
@shahar4499 @jgajek any update here?

jgajek:
The vSphere machinery module already gets the status directly from the vSphere API (see the _status() function in /cuckoo/machinery/vsphere.py).  There must be something else going on -- I'm going to add a call to _status() at the end of each start() and stop() to force the DB status for the machine to be updated.

PR #1675, opened 2017-06-26
Refresh machine power state after each start/stop

jgajek:
As discussed in issue #1348.


jbremer:
What are the possible values of `vm.runtime.powerState`?
What do you think about filtering / normalizing the status values (so that VirtualBox/vSphere/etc represent the same values in the end)?

jgajek:
According to the docs (https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.VirtualMachine.PowerState.html), the possible powerState values are:

poweredOn, poweredOff, suspended

I like the idea of the power state being tracked by Cuckoo in an abstract way, because it doesn't seem like the machinery modules are doing it consistently.

PR #1687, opened 2017-06-30
Error handling for sign.meta.description

stmerry:
Not all my yara rules have a description (apparently). This crashes the summary page if that is the case. That's the way I fixed it on my side for now. You may have a better way, but thought I'd let you know in the meanwhile.

(There are other files too that might need editing)


jbremer:
Thanks! Since you appear to be the first person to run into this issue I've decided to ensure `meta.description` exists from now on (or after reprocessing an older analysis) rather than complicating the printing of `sign.meta.description` (which occurs 5 times throughout the Web Interface). Please also see the commit above. Closing as resolved.

PR #1694, opened 2017-07-04
Add support for URL analysis in cuckoo distributed 

fyhertz:
Usage example: 
`curl http://localhost:9003/api/task -F url="http://www.perdu.com"`

Two commits: one for the small modifications needed in distributed and the other for the migration script since a new field is needed in the task database.


fyhertz:
I have just realized this could be implemented by using the path field for the URL instead of addindg an extra column... What do you think is cleaner? I would not mind to submit another pull request.

jbremer:
Just needs a `category` column for completeness. Pinging @RicoVZ.

PR #1700, opened 2017-07-06
Support for URLs analysis in cuckoo distributed (no db migration needed)

fyhertz:
Usage example:
curl http://localhost:9003/api/task -F url="http://www.perdu.com"

I think this is a better way to implement this feature than my previous pull request as no database migration is needed. The path field is used to store URLs in the task table.

### DIFF
--- a/cuckoo/distributed/api.py
+++ b/cuckoo/distributed/api.py
@@ -37,20 +37,34 @@ def submit_task(url, task):
         enforce_timeout=task["enforce_timeout"],
     )
 
-    # If the file does not exist anymore, ignore it and move on
-    # to the next file.
-    if not os.path.isfile(task["path"]):
-        return task["id"], None
+    # The task is a URL analysis
+    if not task["filename"] and task["path"]:
+        data["url"] = task["path"]
+        try:
+            r = requests.post(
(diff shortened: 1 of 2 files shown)


jbremer:
Although I mostly agree that this PR supersedes #1694, I somewhat also feel like adding a `category` field - much like we have in the Cuckoo Core - would make the changes in this PR slightly less hacky.
What are your thoughts?

fyhertz:
It would be less hacky indeed! In the meantime, how about merging this cheap PR to enable distributed URLs analysis? Another later commit could add the category field and start using it to differentiate samples from URLs (instead of checking if filename is empty) while the path field would still be used for storing the URLs. 

In other words, I think this PR brings distributed closer to a clean solution anyway. What do you think?

jbremer:
If I've learnt one thing then it's the fact that such 'temporary' fixes remain permanent ;-) As we have to maintain it it's much preferable to get it right the first time.
That said, and as you agreed the other option would be slightly better, would you be up for implementing it? Thanks!

jbremer:
I guess the answer is 'no' :-) We'll probably be implementing this ourself eventually (pinging @RicoVZ), closing PR. Feel free to reopen if you'd like to implement the proper approach, as discussed above.

PR #1713, opened 2017-07-14
Fix for Excel ad PowerPoint 2016 to run automatically.

rffuste:
Hi, 

Excel and PowerPoint from Office 2016 did not start automatically.
I’ve updated Office 2016 path for Excel and Powerpoint.

In the same way as Word 2016 has this path: ("ProgramFiles", "Microsoft Office", "root", "Office16", "WINWORD.EXE"),
There is no equivalent path for Excel and PowerPoint.

Without these paths, both applications did not start automatically.
This fix solves the issue.

Regards, 
Rubén


jbremer:
Thanks, excellent find as discussed! Manually merging it and will be pushed soon'ish.

jbremer:
This has been included in `2.0.4a5` (install with `pip install 2.0.4a5`), thanks again!

PR #1715, opened 2017-07-16
Add in parent process name in the process listing for searchability

warewolf:
I'm not great at python, feel free to refactor this.

In short: I use Elastic, and dig through cuckoo reports there, and without this data it's difficult to tie-together a process hierarchy in a query there.

### DIFF
--- a/cuckoo/processing/behavior.py
+++ b/cuckoo/processing/behavior.py
@@ -156,6 +156,11 @@ class GenericBehavior(BehaviorHandler):
             for key, value in process["summary"].items():
                 process["summary"][key] = list(value)
 
+            # having parent process name in the report is incredibly useful in ElasticSearch
+            for item in self.processes.values():
+                if item["process_id"] == process["parent_id"]:
+                    process["parent_name"] = item["process_name"]
+
         return self.processes.values()
 
 class ApiStats(BehaviorHandler):


jbremer:
Looks fine to me. Could you provide us with a use-case for this functionality, though? :)

swackhamer:
This ultimately is a question of schema, and where custom tailored reporting schema logic goes.

Adding in information like full process path, and even duplicating a lot of data just for ES search-ability will add a lot for the Elastic component in general.

potentially putting it here as well:
https://github.com/warewolf/cuckoo/blob/d429f9293a00e0bacc0ce738d067d0c907bc23c2/cuckoo/processing/behavior.py#L103

warewolf:
You wanted a use case, this is a little quick and dirty, but with a proper elastic mapping (e.g. nesting), this sort of thing can happen - dig through elastic for cryptolocker malware, based on dropped filename.
```
{
  "query": {
    "bool": {
      "must": [
        { "term": { "_type": "calls" } },
        { "term": { "calls.category": "filesystem" } },
        { "term": { "calls.api": "ntcreatefile" } },
        { "term": { "calls.arguments.name": "filename" } },
        { "regexp": { "calls.arguments.value": ".*help.*" } }
      ]
    }
  }
}
```

PR #1719, opened 2017-07-17
Add current index into a cuckoo alias for searchability

warewolf:
Add the datestamped cuckoo indexes to a 'cuckoo' alias, so stuff like

`GET http://elastic.cuckoo/cuckoo/_search?q=some.json.path:value`

hits all the indexes, and you don't have to specify them.

### DIFF
--- a/cuckoo/reporting/elasticsearch.py
+++ b/cuckoo/reporting/elasticsearch.py
@@ -185,6 +185,9 @@ class ElasticSearch(Report):
         }[elastic.index_time_pattern])
         self.dated_index = "%s-%s" % (elastic.index, date_index)
 
+        # add this index to the cuckoo alias for searching
+        elastic.client.indices.put_alias(index=dated_index, name='cuckoo')
+
         # Index target information, the behavioral summary, and
         # VirusTotal results.
         doc = {


jbremer:
Sounds good. I'm fine with merging this, but just asking some others - @KillerInstinct @razuz @RicoVZ @swackhamer.

RicoVZ:
Good idea, looks fine to me!

swackhamer:
Why not just index wild card the search?
```
GET http://elastic.cuckoo/cuckoo/_search?q=some.json.path:value
->
GET http://elastic.cuckoo/cuckoo*/_search?q=some.json.path:value
# example dated search:
GET http://elastic.cuckoo/cuckoo-2017-03*/_search?q=some.json.path:value
```

Index aliases can be used to map 1 alias to 1 index, usually to control which index a indexer goes into.
```
index alias: cuckoo -> index: cuckoo-000001
```
Then rotate that index efficiently with something like Curator: https://www.elastic.co/guide/en/elasticsearch/client/curator/current/ex_rollover.html


Index aliasing like this I think should be done with something like curator:
https://www.elastic.co/guide/en/elasticsearch/client/curator/current/ex_alias.html

PR #1743, opened 2017-08-01
inetsim with dummy port support

doomedraven:
* for linux samples is useful to be able activate dummy port in `inetsim
is 1 tcp/udp` and in `fakenet-ng is tcp/udp 1337`
* also added support for correct ports forwarding if dummy port
activated


doomedraven:
just saw that is already integrated, closing

PR #1745, opened 2017-08-01
add check_command_line

doomedraven:
instead of do
```
all/any([re.match(pattern, line) for line in
self.get_summary("command_line")])
```


jbremer:
Thanks, manually merged up.

Issue #1752, opened 2017-08-02
Error parsing command-line


doomedraven:
```
2017-08-01 15:38:52,648 [cuckoo.common.scripting] WARNING: Error parsing command-line: cmd /c ""C:\Users\XXX\AppData\Local\Temp\4692\FEC9.bat" "C:\Users\XXX\AppData\Roaming\MICROS~1\Authspex\bitspnet.exe" "C:\Users\XXX\AppData\Local\Temp\URSNIF~1.EXE""
2017-08-01 15:38:52,655 [cuckoo.common.scripting] WARNING: Error parsing command-line: cmd  /C ""C:\Users\XXX\AppData\Roaming\MICROS~1\Authspex\bitspnet.exe" "C:\Users\XXX\AppData\Local\Temp\URSNIF~1.EXE""

2017-08-02 14:59:29,090 [cuckoo.common.scripting] WARNING: Error parsing command-line: cmd  /C ""C:\Users\XXX\AppData\Roaming\MICROS~1\Authspex\bitspnet.exe" "C:\Users\XXX\AppData\Local\Temp\URSNIF~1.E
XE""
2017-08-02 14:59:29,093 [cuckoo.common.scripting] WARNING: Error parsing command-line: cmd  /C ""C:\Users\XXX\AppData\Roaming\MICROS~1\Authspex\bitspnet.exe" "C:\Users\XXX\AppData\Local\Temp\URSNIF~1.E
XE""
2017-08-02 14:59:29,096 [cuckoo.common.scripting] WARNING: Error parsing command-line: cmd  /C ""C:\Users\XXX\AppData\Roaming\MICROS~1\Authspex\bitspnet.exe" "C:\Users\XXX\AppData\Local\Temp\URSNIF~1.E
XE""
2017-08-02 14:59:29,098 [cuckoo.common.scripting] WARNING: Error parsing command-line: cmd  /C ""C:\Users\XXX\AppData\Roaming\MICROS~1\Authspex\bitspnet.exe" "C:\Users\XXX\AppData\Local\Temp\URSNIF~1.E
XE""
2017-08-02 14:59:29,101 [cuckoo.common.scripting] WARNING: Error parsing command-line: cmd  /C ""C:\Users\XXX\AppData\Roaming\MICROS~1\Authspex\bitspnet.exe" "C:\Users\XXX\AppData\Local\Temp\URSNIF~1.E
XE""
[...]

PR #1755, opened 2017-08-04
fix of https://github.com/cuckoosandbox/cuckoo/issues/1752

doomedraven:
````
>>> shlex.split('cmd  /C ""C:\Users\XXXX\AppData\Roaming\authspex\bitspnet.exe" "C:\Users\XXXX\AppData\Local\Temp\686963~1.EXE""', posix=False)

>>> shlex.split('cmd  /C "C:\Users\XXXX\AppData\Roaming\authspex\bitspnet.exe" "C:\Users\XXXX\AppData\Local\Temp\686963~1.EXE"', posix=False)
['cmd', '/C', '"C:\\Users\\XXXX\\AppData\\Roaming\x07uthspex\x08itspnet.exe"', '"C:\\Users\\XXXX\\AppData\\Local\\Temp\x0686963~1.EXE"']
>>>
```

### DIFF
--- a/cuckoo/common/scripting.py
+++ b/cuckoo/common/scripting.py
@@ -22,6 +22,7 @@ class Scripting(object):
         if isinstance(cmdline, (tuple, list)):
             return cmdline
         try:
+            cmdline = cmdline.replace('""', '"')
             return shlex.split(cmdline, posix=False)
         except ValueError:
             log.warning("Error parsing command-line: %s", cmdline)


jbremer:
This is not entirely accurate, actually. Just tested and `python.exe a.py ""a" "b" "c""` with `a.py` being `import sys; print sys.argv` prints out `['a.py', 'a b c']`. I do agree that we should be improving this, though.

jbremer:
Closing PR as this isn't the correct way to approach the issue (see also my comment above), thanks though.

Issue #1396, opened 2017-04-12
Unable to communicate with cuckoo 2.0.0 client


thebellcup:
Hi,

My host cuckoo is unable to communicate with the new client agent with the following error:

2017-04-12 11:33:08,922 [cuckoo.core.startup] WARNING: Unable to import yara (please compile from sources)
2017-04-12 11:33:08,922 [cuckoo.core.resultserver] DEBUG: ResultServer running on 192.168.163.129:2042.
2017-04-12 11:33:08,924 [cuckoo.core.scheduler] INFO: Using "physical" as machine manager
2017-04-12 11:33:09,015 [cuckoo.machinery.physical] DEBUG: Getting status for machine: physical1.
2017-04-12 11:33:09,025 [cuckoo.machinery.physical] DEBUG: Received unknown exception: <ProtocolError for 192.168.163.200:8000/RPC2: 404 Not Found>.
CuckooCriticalError: Error initializing machines: Unknown error occurred trying to obtain the status of physical machine physical1. Please turn it on and check the Cuckoo Agent.
(venv) aaa@ubuntu:~$ 

My current setup:
Host machine ip 192.168.163.129 (VM)
Running on Ubunbtu 16.04
It was setup following https://cuckoo.sh/docs/installation/host/requirements.html

Guest machine ip 192.168.163.200 (VM)
Running on Windows 7 32bit sp1
Installed with python 2.7.13, agent.py in startup folder and listening on port 8000

They are configured to talk to each other using physical machinery

Any help is greatly appreciated. Thanks!

SparkyNZL:
a couple of things you can do is

Ping that IP address from your cuckoo to make sure it can be seen, if it cant please review the networking for the BMH and make sure youi have turned off the Windows FW etc.

telnet to your BMH  again if this isnt working then you have a network issue, make sure your windoze FW is turned off 

once you have verified that the comms is working in both directions let us know.

CHeers.

thebellcup:
Thanks SparkyNZL.

Both host and guest machine can ping each other. 
Guest (windows) firewall had been turned off.
Host (ubuntu) iptables accept all traffic.
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT

Telnet from host to guest machine port 8000 is accepted.
telnet 192.168.163.200 8000
Trying 192.168.163.200...
Connected to 192.168.163.200.
Escape character is '^]'.

Network detail of each machine:
[...]

PR #1759, opened 2017-08-07
Update requirements.txt -

jmlynch:
Alert users that are following the documentation for installs that include physical machinery to the fact that the new cuckoo agent does not currently support physical machinery & let them know to use old agent. https://github.com/cuckoosandbox/cuckoo/issues/1396


jbremer:
Sweet, thanks! Manually merged this PR. Closing as resolved.

PR #1770, opened 2017-08-11
Linux analysis round #2

daanfs:
- Remove a bunch of dead/deprecated code like the problkm and handle pid following in the systemtap parser. 

- Remove strace fallback since it was unused and didn't conform to the systemtap log format. Payload is only ran once.

- Remove magic 10 second sleep from systemtap handler, behavior analysis finishes once systemtap is completed.

- Set `ps`-like command line info on successful execve call

- Some systemtap parsing improvements (successful execve calls are now properly logged, return status on error is properly displayed, structs are parsed as a single argument)

- Specify ubuntu 17.04 as recommended platform due to latest systemtap version.


jbremer:
Thanks. I've pushed this to the `linux2` branch. As my last commit mentions a few more unit tests are required before merging this completely. Closing PR as resolved, though.

PR #1773, opened 2017-08-14
enlarge shellcode size limit, and wait while shellcode is executing.

shellbombs:
2048 is too small for shellcode, and shellcode may create threads, if return immediately, the threads have no opportunity to execute. eg: the Xshell backdoor event happened recently. the Xshell backdoor shellcode.


shellbombs:
sorry, i made a small mistake, i will resend the patch.

PR #1774, opened 2017-08-14
enlarge shellcode size limit, and wait while shellcode is running

shellbombs:
2048 is too small for shellcode, and shellcode may create threads, if return immediately, the threads have no opportunity to execute. eg: the Xshell backdoor event happened recently. the Xshell backdoor shellcode.


jbremer:
Would be more useful if you'd also update the compiled binary :-) As-is, nothing will change in production!

shellbombs:
binary updated. :-) 
https://github.com/cuckoosandbox/cuckoo/pull/1774#issuecomment-324664959

jbremer:
Additionally, we should probably also have both 32-bit and 64-bit shellcode support.

jbremer:
Merged, thanks! No 64-bit support for now, though.

PR #1779, opened 2017-08-15
allow some invalid urls with double schema for example

doomedraven:
maybe will be extended in future


jbremer:
Manually merged, thanks!

Issue #1485, opened 2017-04-26
No MISP expression code in Django


hakawati:
Hi,

It works fine by enabling misp in `reporting.conf` and `processing.conf`. 
- reporting.conf
> [misp]
> enabled = yes
> url = http://192.168.0.250
> apikey = JR520ltb3SxStCbuluMFHRGP4WcUz4FW8sNd5rUo
> mode = maldoc ipaddr hashes url

- processing.conf
> [misp]
> enabled = yes
> url = http://192.168.0.250
> apikey = JR520ltb3SxStCbuluMFHRGP4WcUz4FW8sNd5rUo

[nav-sidebar.html](https://github.com/cuckoosandbox/cuckoo/blob/master/cuckoo/web/templates/analysis/pages/nav-sidebar.html) contains MISP-related content. 

> 	{% if report.analysis.misp %}
> 			\<li>
> 				\<a href="{% url "analysis" report.analysis.info.id "misp" %}"
> 				{% if env.view_kwargs.page == 'misp' %} class="active" {% endif %}>
> 					\<i class="fa fa-flag"></i> 
> 					\<span>MISP</span>
> 					\<em>{{ report.analysis.misp|length }}</em>
> 				\</a>
> 			\</li>            
>         {% endif %}

However, the sidebar of the Django web interface has nothing to do with MISP. Thanks for reading this issue.

Roslekas:
Hi, i noticed the same, MISP sidebar does not appear in Django web interface, with Debugging mode shows that Executed proccesing module "MISP" and Executed reporting module "MISP"...

![image](https://cloud.githubusercontent.com/assets/16913780/25429583/303afbcc-2a82-11e7-827a-a109e98dca21.png)

![image](https://cloud.githubusercontent.com/assets/16913780/25429572/24e9395a-2a82-11e7-8b51-7e928302b42e.png)

Thanks

SparkyNZL:
same as #1460 ?

SparkyNZL:
Look like the MISP processing is broken in 2.0.3a2. Im going to open a new TIcket for this,

jbremer:
Thanks, bit late to the party but this should now be resolved due to #1784 #1789. Closing issue!

Issue #1460, opened 2017-04-23
Cuckoo Package : WGUI, MISP page


SparkyNZL:
Hi guys,

there is no scroll bars on the MISP page and the Left hand menu is all screwed up :(

![screenshot from 2017-04-23 13-29-19](https://cloud.githubusercontent.com/assets/17818264/25309737/f4743168-2828-11e7-9e30-c637bfbdc2f5.png)

jbremer:
Ah, thanks. We'll have to check that out.

jbremer:
Supposedly fixed in #1784 #1789. Thanks for reporting. Closing as resolved.

PR #1784, opened 2017-08-16
fixed old misp screen

tatsui:
Replace new sidebar and add footer.

PR #1789, opened 2017-08-19
fix broken misp page

razuz:
This patch fixes broken MISP page and adds fluid nav bar to it
from what I see it's related to #1460 and probably #1485


jbremer:
Duplicate of #1784. Merging that one instead - it's slightly easier due to not having indentation changes.

Issue #1785, opened 2017-08-16
Add www.msftncsi.com to whitelist


chrisdoman:
Hi,

 I'm getting lots of traffic to the MS FQDN www.msftncsi.com on win7 hosts.
 Other subdomains are whitelisted but not that one.

Thanks,

Nwinternights:
You Can Simply disabile Active probing on guests HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\services\NlaSvc\Parameters\Internet
Double-click on EnableActiveProbing and change its value from 1 to 0. Reboot  and then take a snapshot

SparkyNZL:
or if you didnt do this in a load of vms (cough) you can just add it to the whilelist called domain.txt which is located in /usr/local/lib/python2.7/dist-packages/cuckoo/data-private/whitelist

keep in mind when you upgrade cuckoo you will need to re-edit this and add the domains again, i tend to keep another copy which i just copy back after an update :)

chrisdoman:
Yup I added it to domains.txt - thanks :) 

Might want to add it to domains.txt as standard for everyone, or disable it as part of the vmcloak standard build.

jbremer:
I've added it to the whitelist. Closing issue as resolved. Thanks!

Issue #1716, opened 2017-07-17
Always the same signatures detected?


Roslekas:
Hi,

always getting same signatures, how to disable or change severity ? because i manually disabled ,for example, firewall and Host detects as suspicious activity. Maybe this is error ?

![image](https://user-images.githubusercontent.com/16913780/28257189-844c90b0-6ad1-11e7-916b-9ed888033642.png)

Thanks

Hullgj:
You can change the signatures. Refer to here:
http://docs.cuckoosandbox.org/en/latest/customization/signatures/

Roslekas:
Hi again,

also getting same Potential malicious URL's, seems it not malicious, tried to whitelist it in network.py, but it didn't work. Where to find instruction how whitelist URL's and dead_host's, IP's?
Thanks

![image](https://user-images.githubusercontent.com/16913780/28875395-1a6e6d88-779e-11e7-859e-c81f2e980842.png)
![image](https://user-images.githubusercontent.com/16913780/28875419-2a5ac944-779e-11e7-96fe-7b68ba3b2347.png)

jbremer:
Feel free to compile a list of URLs or domains that should be whitelisted, @Roslekas!

Issue #1465, opened 2017-04-24
Cuckoo Package: Upgrade causes whitelist overwrite


SparkyNZL:
Hey guys, when you do an upgrade or update with pip, this deletes the domain.txt white list, which is a pain as i have quite a big white list and i have to keep updating it manually, can the while list be put in the .cuckoo directory ?

jbremer:
Hm, I had moved it to the private directory because I thought that'd be better - assuming users don't update it all that much (see also f240c5eb4d6abf004ed15917878fc8208df682c8), but perhaps we can have Cuckoo load both files; from private and from CWD.

razuz:
@jbremer you want to make this configurable or just load both files in a row ?

another thing I noticed is that whitelists are loaded even if whitelisting is not enabled. maybe that check should be moved into _build_whitelist() instead of _is_whitelisted()

razuz:
just for the ref ... same thing applies here
https://github.com/cuckoosandbox/cuckoo/blob/master/cuckoo/common/whitelist.py#L13

jbremer:
This will indeed be possible in the upcoming release. Closing as resolved.

PR #1791, opened 2017-08-20
Add ability to include whitelist from $CWD aswell

razuz:
This PR brings back option to add second custom whitelist via path $CWD/whitelist/domain.txt when available and merges it with built in whitelist from package

Should resolve issues one way or another #1785 #1716 #1465 #948


jbremer:
Manually merged, closing it! Thanks.

PR #1794, opened 2017-08-21
better explain misleading cuckoo init failure

razuz:
So when cuckoo init fails due to existing dir then this error message is a bit misleading. Added another note there which could save a few hours for newbies :)


jbremer:
Merged up - thanks! Closing.

PR #1811, opened 2017-08-29
Fix coverage of LinuxSystemTap tests

daanfs:
Additional behaviour also tested now.


jbremer:
Good stuff, all merged now!

PR #1812, opened 2017-08-30
Add an argument to query a report in the API.

tUn4:
Hi there,
I have approximately 20k sample to analyze, once i get the report of analysis, it's pretty large.
And may be i don't need all the information in the report, therefore i added an option to the API to get a  specific field in the report result, such as "info", "signatures", "network", ...
I think it's useful when we have a lot of data, sample.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,13 @@ before_install:
         brew update || brew update
         brew install libmagic cairo pango mongodb
         brew services start mongodb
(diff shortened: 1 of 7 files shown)


jbremer:
Not optimal with regards to memory usage as well as API usage (e.g., what if you wanted to select multiple keys or a subkey?), but a step forward regardless, so merged it anyway. Thanks!

PR #1816, opened 2017-09-01
Correct typo

mruessler:
The variable $night_table_odd_color has a typo


jbremer:
Thanks. I've simply removed the two variables as they weren't being used in the first place. Closing as resolved.

Issue #1828, opened 2017-09-08
Development with the Python Package


rainkin1993:
Hi I want to clone the cuckoo project and do some developments.

I clone the git by this command: 
`https://github.com/cuckoosandbox/cuckoo.git`

After running the commands in the following tutorial:
http://docs.cuckoosandbox.org/en/latest/development/package/

I run the command `cuckoo -d` and get this error:

IOError: [Errno 2] No such file or directory: '/home/myname/Documents/cuckoo/cuckoo/data-private/.cwd'

And I try to install cuckoo from source code by running this command `python setup.py install`
But I still get the same error.

Thanks

jbremer:
You're correct - I had to change this for the `2.0.4` release as there were some issues popping up and didn't think of this. Will see if there's a better workaround, but for now you can run `python setup.py sdist install` or `python setup.py sdist develop` and it should work. Perhaps that's fine as-is to be included in our documentation.

jbremer:
Updated docs as stated above - will have to do for now. Thanks. Closing as resolved.

PR #1829, opened 2017-09-08
Removed ifconfig from rooter, as ip can deliver the same information

zaddach:
This pull request removes the `ifconfig` utility from the rooter function. The `ip` iutility is required anyways and can do the same job, and `ifconfig` is redundant (and not present on my system, whereas `ip` is).


zaddach:
The Travis build fails due to #1828

jbremer:
Thanks! Cherry-picked it and included the #1828-related "fix" as well :-)

PR #1855, opened 2017-09-20
Use iptables insert action (-I) in forward_enable

seanthegeek:
An unknown Ubuntu security update (probably `libvirt`) got installed automatically by `unattended-updates`, and changed the way my iptables `FORWARD` chain is configured at boot:

```
Chain FORWARD (policy DROP)
target     prot opt source               destination
ACCEPT     all  --  anywhere             192.168.101.0/24
ACCEPT     all  --  192.168.101.0/24     anywhere
ACCEPT     all  --  anywhere             anywhere
REJECT     all  --  anywhere             anywhere             reject-with icmp-port-unreachable
REJECT     all  --  anywhere             anywhere             reject-with icmp-port-unreachable
ACCEPT     all  --  anywhere             anywhere
REJECT     all  --  anywhere             anywhere             reject-with icmp-port-unreachable
REJECT     all  --  anywhere             anywhere             reject-with icmp-port-unreachable
ufw-before-logging-forward  all  --  anywhere             anywhere
ufw-before-forward  all  --  anywhere             anywhere
ufw-after-forward  all  --  anywhere             anywhere
ufw-after-logging-forward  all  --  anywhere             anywhere
ufw-reject-forward  all  --  anywhere             anywhere
ufw-track-forward  all  --  anywhere             anywhere
```

Currently,  any rules added by `rooter` were appended (`-A)` to the bottom of the chain. Because of the `REJECT     all  --  anywhere             anywhere             reject-with icmp-port-unreachable` rule is above, all the traffic that should go from a VM through a VPN is rejected with `Destination port unreachable`.

This commit fixes this problem by having rooter use the insert (`-I`) action, adding the rules to the top of the forward chain.

This is a similar problem and solution as I described  in PR #1621. That PR was rejected due to the unnecessary use of the `-I` option  in other iptables commands. This PR only changes the `forward_enable` command.

### DIFF
--- a/cuckoo/apps/rooter.py
+++ b/cuckoo/apps/rooter.py
@@ -152,10 +152,10 @@ def dns_forward(action, vm_ip, dns_ip, dns_port="53"):
 def forward_enable(src, dst, ipaddr):
     """Enable forwarding a specific IP address from one interface into
     another."""
-    run(s.iptables, "-A", "FORWARD", "-i", src, "-o", dst,
+    run(s.iptables, "-I", "FORWARD", "-i", src, "-o", dst,
         "--source", ipaddr, "-j", "ACCEPT")
 
-    run(s.iptables, "-A", "FORWARD", "-i", dst, "-o", src,
+    run(s.iptables, "-I", "FORWARD", "-i", dst, "-o", src,
         "--destination", ipaddr, "-j", "ACCEPT")
 
 def forward_disable(src, dst, ipaddr):


doomedraven:
which version of ubuntu do you use? i have 16.04.x(with all upgrades) and don't have that iptable rule

seanthegeek:
@doomedraven 17.04. I wish there way a way to see what created this rule. It's really weird that it's in there four times. 

Having the `FORWARD` rule inserted doesn't harm anything, and works around edge cases like this.

doomedraven:
yes yes, i just was curious which version of ubuntu do that, long live LTS :D

seanthegeek:
Closing this ancient PR

PR #1883, opened 2017-10-11
Launching the web interface faster and in a safer way.

shahar4499:
By default, django's web server forks itself to create another process that constantly scans the source code for changes, and than restarts the server.
This sometimes makes the launching of the web interface very slow + has a small potential to cause loss of state.
Passing a `--noreload` argument to django prevents this behavior.

### DIFF
--- a/cuckoo/main.py
+++ b/cuckoo/main.py
@@ -545,7 +545,7 @@ def web(ctx, args, host, port, uwsgi, nginx):
 
     try:
         execute_from_command_line(
-            ("cuckoo", "runserver", "%s:%d" % (host, port))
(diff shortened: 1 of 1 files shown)


shahar4499:
No idea why the code coverage was affected so much by this small change, is this really accurate?

jbremer:
Thanks! Manually merged this and only enabled this when running, e.g., `cuckoo web`, as opposed to `cuckoo -d web`, in which case the previous mode (i.e., no `--noreload`) is used. Naturally we still have to debug the Web Interface on a day-to-day basis ;-)

jbremer:
Unmerged this as it did in fact break some unit tests (likely due to the `os.chdir()` in `web()` in `cuckoo/main.py`).

PR #1887, opened 2017-10-12
Fix fog authentication

martinmesser:
Authentication while using cuckoo with fog 1.4.4 failed because of changes in fog website design.
Add missing login parameters to authenticate user, and modify request links.
With this PR we parse the needed information out of the fog site content.


jbremer:
Thanks for your contribution! Closing this as duplicate of #1288. I've changed the hard error (from #1288) to a warning in case Fog isn't running either `1.3.4` or `1.4.4`, which I guess should work for now. Please let me know if you run into any issues!

PR #1888, opened 2017-10-13
Making it harder to detect Cuckoo by click events

G4lB1t:
The click events in _click_mouse()_ can be easily detected (see https://github.com/G4lB1t/SmoothCriminal) so I suggest adding some randomness to it.

### DIFF
--- a/cuckoo/data/analyzer/windows/modules/auxiliary/human.py
+++ b/cuckoo/data/analyzer/windows/modules/auxiliary/human.py
@@ -121,10 +121,10 @@ def move_mouse():
 
 def click_mouse():
     # Move mouse to top-middle position.
-    USER32.SetCursorPos(RESOLUTION["x"] / 2, 0)
+    USER32.SetCursorPos(RESOLUTION["x"] / random.choice([2, 3, 4]), random.choice([0, RESOLUTION["y"]]))
     # Mouse down.
     USER32.mouse_event(2, 0, 0, 0, None)
-    KERNEL32.Sleep(50)
+    KERNEL32.Sleep(random.choice([20, 30, 40, 50])
     # Mouse up.
     USER32.mouse_event(4, 0, 0, 0, None)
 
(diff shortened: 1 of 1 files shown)


kevross33:
Hi,

Great stuff. I am wondering if some other ideas and I would really appreciate if you would please consider some of them if they were at all possible. 

- Slowing down the mouse a bit so it is more human speed as some malware measures speed mouse is getting about and if a human could move that fast.

- introduce scrolling in documents (PDFs, office DOCs etc.) I have seen some malware documents and macros this year which only activate when scrolled down to next page.

- Clicking on any embedded buttons (common to embed for instance office in PDFs with a button which opens it when clicked).

- Accelerating system time appropriately when skipping sleep and stalling instructions (i.e if I check system time and it says 1:20 and I sleep 10 mins and I check time again and it is 1:20:01 I know I have been accelerated and this has been used by malware).

References:
https://www.fireeye.com/content/dam/fireeye-www/current-threats/pdfs/pf/file/fireeye-hot-knives-through-butter.pdf
https://www.lastline.com/labsblog/not-so-fast-my-friend-using-inverted-timing-attacks-to-bypass-dynamic-analysis/

G4lB1t:
@kevross33 I have some code implementing evasion techniques specifically against the smoothness of mouse movement (see the link to SmoothCriminal) but I already wrote a "smooth mouse moving" script to counter it - hopefully I'll fix some glitches it has and will push it soon as well.

cuckoo/data/analyzer/windows/modules/auxiliary/human.py
```diff
@@ -120,11 +120,11 @@ def move_mouse():
...
 
 def click_mouse():
-    # Move mouse to top-middle position.
-    USER32.SetCursorPos(RESOLUTION["x"] / 2, 0)
+    # Move mouse to either top or bottom of the screen, with a bit of randomness
+    USER32.SetCursorPos(RESOLUTION["x"] / random.choice([2, 4, 8]), random.choice([0, RESOLUTION["y"]]))
     # Mouse down.
     USER32.mouse_event(2, 0, 0, 0, None)
-    KERNEL32.Sleep(50)
+    KERNEL32.Sleep(random.choice([20, 30, 40, 50])
```
razuz: FYI,
missing one ) here in the end

Issue #426, opened 2014-12-15
KVM VM does not start


stevengoossensB:
Hi,

I have a problem with the default libvirt implementation and the override ivm.py is doing. It seems that the method to restore the snapshot in the libvirt API does not automatically start the VM after restoring. This creates the issue that cuckoo is restoring the snapshot, but is not starting the VM. When I manually start the VM while Cuckoo is waiting for it, everything after that goes well.

I guess a kvm.py implementation would resolve this issue. Does anyone have one yet?

Rgds,

Steven

voleg:
+1

voleg:
temporary workaround:

```python
369             try:
370                 self.vms[label].revertToSnapshot(snapshot, flags=0)
371 
372                 # look for bug status plz..
373                 # https://github.com/cuckoobox/cuckoo/issues/426
374                 self.vms[label].create()
375 
376             except libvirt.libvirtError:
377                 raise CuckooMachineError("Unable to restore snapshot on "
378                                          "virtual machine {0}".format(label))
```

jekil:
Thanks for reporting, i am investigating on it.

jekil:
I run some tests with latest Xubuntu and i am unable to reproduce the bug. Can you please add more details? (OS, libvirt version, etc.)

KillerInstinct:
If I'm reading your issue correctly, it seems to have shut your VM down before taking the snapshot. The snapshot should be taken when the VM is On and the agent is listening, per the documentation.

rep:
Let me be the cleaner on this one. We can't reproduce and we don't have enough information anyways - this might have been a temporary glitch in the libvirt lib or our module for it - which has since been fixed.

Closing this, feel free to re-open if it happens again. With more details then, please :)

bbb31:
I want to confirm; this *is* an issue with Ubuntu Server 16.04. Cuckoo hangs on attempting to start the VM; giving the following error:

```
...
2017-10-18 20:52:45,710 [cuckoo.common.abstracts] DEBUG: Waiting 61 cuckooseconds for machine win7-cuckoo to switch to status ('running',)
2017-10-18 20:52:45,717 [cuckoo.core.scheduler] ERROR: Error starting Virtual Machine! VM: win7-cuckoo, error: Timeout hit while for machine win7-cuckoo to change status
```

Other info:

```
$ virsh -v
1.3.1

$ libvirtd -V
libvirtd (libvirt) 1.3.1

$ lsb_release -a
[...]

PR #1904, opened 2017-10-19
start kvm domain

bbb31:
There is an old bug reported; #426 that I ran into; I noticed that the domain simply isn't started. This PR starts the domain.

I have tested it to be working on my end. See the comments in #426 where I provide more info.

### DIFF
--- a/cuckoo/common/abstracts.py
+++ b/cuckoo/common/abstracts.py
@@ -365,6 +365,7 @@ class LibVirtMachinery(Machinery):
                 vm = self.vms[label]
                 snapshot = vm.snapshotLookupByName(vm_info.snapshot, flags=0)
                 self.vms[label].revertToSnapshot(snapshot, flags=0)
+                self.vms[label].create()
             except libvirt.libvirtError:
                 msg = "Unable to restore snapshot {0} on " \
                       "virtual machine {1}".format(vm_info.snapshot, label)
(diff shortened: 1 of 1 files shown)


doomedraven:
that is just bcz you not follow the documentation, and as in another issue where you wrote, there you have the answer https://github.com/cuckoosandbox/cuckoo/issues/426#issuecomment-72355653

bbb31:
What part of the documentation? I've read through it; so much so that I did not find *why* that I decided to read the source and noticed that the VM was not starting.

doomedraven:
i pointed to the comment with answer to your quesiton, all vms should be in running state when you taking the snapshot

bbb31:
Alright, it does work so long as the snapshot is in "running" state. However, this shouldn't be a requirement. If the specific snapshot is provided, the code should check if the VM is running after; and if not start it regardless.

doomedraven:
not make sense waste time in OS boot/shutdown, there also is an antivm if idle time is low

bbb31:
It should probe the state of the specified VM; if it's off attempt to revert to the specified snapshot; then power it on if the snapshot that was restored to did not result in the VM automatically starting. What I'm saying is; you shouldn't rely on an arbitrary state of a snapshot to properly conduct an analysis. You should be able to disable the firewall, automatic updates, and put the agent in the startup folder; take a "clean" snapshot and just revert to that.

As for antivm; it's no diff than taking a snapshot after 30 sec of the VM running vs starting up and then running the malware a few seconds after start. Instead, a warning should be provided regardless.

daanfs:
As the documentation currently states that the snapshot should be in a running state, this seems to be functioning as intended.

@bbb31: Is there any benefit to using a snapshot that is not in a running state?

PR #1921, opened 2017-10-29
Remove artifacts from older installations

thorkill:
It seems that the .cwd file is no more created and without it the tests should fail.


thorkill:
Ok, it looks like travis does do everything fine.

Issue #1119, opened 2016-10-03
Error Running certain EXE files


fez219:
Hello,

I'm just starting to actually test my sandbox with real malware. I'm also running into issues.

So far, I've been able to run safe .exe files through the sandbox with no issues. However, when I try to run an actual malware sample (also a .exe), I get a stranger error.

My Windows 7 guest gives me this error message:

Inject error

[-] Error adding task to APC queue: %Id

I can't find the solution to this error anywhere. Help would be greatly appreciated!

Thanks!

-Alex

doomedraven:
can you provide some log/hashes, more details

fez219:
Please note that my sandbox works fine with PAfish and the installer for Notepad++. I also have my sandbox disconnected from the internet on purpose.

I'm trying to run malware with these hashes:

- MD5: bb72bed531d75dadb3587b2a1ec10da2
- SHA1: f450ebc7273d0dc1b6cebf9360ebcdbfec318d57
- SHA256: 5038e948fda799bc6b779c1c54d6a72c03a4e571615f7a5529eb5549dfb21da2

I just ran it again beginning the command with free=yes, and did not get the error. Here's the analysis.log:

2016-10-05 10:15:20,000 [analyzer] DEBUG: Starting analyzer from: C:\ekmzslko
2016-10-05 10:15:20,015 [analyzer] DEBUG: Pipe server name: \\.\PIPE\WsRepawyvCeYvCWVXKgKUKqKeXrS
2016-10-05 10:15:20,015 [analyzer] DEBUG: Log pipe server name: \\.\PIPE\jtpzPRuVbIyIfgRnGdaI
2016-10-05 10:15:20,015 [analyzer] DEBUG: No analysis package specified, trying to detect it automagically.
2016-10-05 10:15:20,015 [analyzer] INFO: Automatically selected analysis package "exe"
2016-10-05 10:15:20,890 [analyzer] DEBUG: Started auxiliary module Disguise
2016-10-05 10:15:23,233 [analyzer] DEBUG: Loaded monitor into process with pid 480
2016-10-05 10:15:23,233 [lib.api.process] INFO: Successfully injected process with pid None
2016-10-05 10:15:23,233 [analyzer] DEBUG: Started auxiliary module DumpTLSMasterSecrets
2016-10-05 10:15:23,233 [analyzer] DEBUG: Started auxiliary module Human
[...]

PR #1947, opened 2017-11-11
fix injection issue with .NET samples for AnyCPU target

daanfs:
This fixes the issue described in https://github.com/cuckoosandbox/cuckoo/issues/1119


jbremer:
Thanks! Merged manually.

PR #1966, opened 2017-11-17
fix drop rule in router

doomedraven:


jbremer:
Manually applied changes. Thanks!

Issue #1959, opened 2017-11-15
Score doesn't get updated in the recent list after re-processing


sebdg:
Hi all, 

I've analyzed some projects a first time without suricata or snort. 
These are shown in the recent list in the web ui.
When I re-process with snort and suricata, alerts are raised and visible.
The score on the report page itself went from 1 to 2.2, but in the recent list this remains 1.
I'm using 2.0.4. No errors nor warnings during the  re-processing command exection.

Kind Regards

sebdg:
@jbremer you can close this issue as it is about to be merged

jbremer:
You can do that yourself too, btw, but yes, thanks!

PR #1985, opened 2017-11-28
Fix display issues after reprocess

sebdg:
There is an update issue on the recent list, data is not correct after reprocessing.
The data is sorted by insertion id and not report id, this is then amplified with a grouping loop lower.
The score is not updated for example.

### DIFF
--- a/cuckoo/web/analysis/views.py
+++ b/cuckoo/web/analysis/views.py
@@ -57,8 +57,7 @@ def chunk(request, task_id, pid, pagenum):
         {
             "behavior.processes.pid": 1,
             "behavior.processes.calls": 1
-        },
-        sort=[("_id", pymongo.DESCENDING)]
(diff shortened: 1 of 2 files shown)


sebdg:
fixes issue #1959

sebdg:
There is a similar issue in the Behavior tab, the calls displayed are not from the latest report.
Same goes for searching and filtering view of that tab.

jbremer:
I've manually merged this, thanks! Will close issue once it's been shipped in an official release.
Note that I've skipped the `remove extra line breaks` commit as I consider the current syntax as just fine.

jbremer:
This has been resolved in our `2.0.5` release which we put out just now. Please upgrade and try again! Closing issue, but feel free to reopen if the issue(s) persist for you. Thanks!
Please find also our blogpost https://cuckoosandbox.org/blog/205-office-dde

PR #1990, opened 2017-12-02
qemu + qemu antivm fix

doomedraven:
+ one antivm, allow specify -cpu to avoid `QEMU virtual machine` as cpu
@daanfs @jbremer 

could you keep the rest of the arches "undocumented"so they kinda not supported but inside of the file so we can use vms for networking at least and static part, plz


doomedraven:
btw arm changed as ubuntu 17.04 for arm requires that settings

PR #2001, opened 2017-12-06
Fix resubmit

### LINKED ISSUE #1370: Cuckoo Package : Dropped file resubmit results in ERROR!

### LINKED ISSUE #1644: Sample resubmition stop working after some period of time
I'm not  sure why, but after an undetermined period of time, clicking the resubmit sample button on a completed task will return a page that says the sample does not exist, but the download button works, so the sample clearly does still exist.

sebdg:
This fixes an issue in resubmit.
The filepath taken is the temporary file used to submit.
Once analyized another copy is held in the analysis folder. 
That one remains available to download.

Fix #1644 for sure
Fix #1370 also,


jbremer:

PR #2003, opened 2017-12-07
elasticsearch and mongodb reporting avoid doubles

sebdg:
This fix ensures that previous analysis results get deleted before saving new items.
After applying one can reprocess its reports to clean existing duplicates.


sebdg:
if someone using procmon can test and reprocess an analysis to make sure it works there too.

PR #2031, opened 2017-12-28
Fix for JSE files

ALange:
Fix for processing JSE files (JSE != JS)


jbremer:
Thanks! Manually merged it.

PR #2039, opened 2018-01-02
Detect compatible browsers.

pravic:
Treat Chromium compatible (Chrome, Opera, Vivaldi, etc.) or Firefox-based browsers as supported ones.

It does not check browser versions though, so it will be confusing in case of the old Opera for example (`<= 12.x`).


pravic:
Hello? Is anybody here?

doomedraven:
i wouldn't expect merge before next release

doomedraven:
hehe another dead pr ;)

PR #2042, opened 2018-01-03
Added Curtain module

karttoon:
Curtain is a new module I'm submitting which takes PowerShell ScriptBlock logs off of a Windows guest (if enabled), parses them, cleans them up, and displays them to a user. It's primary purpose is to help analyst navigate obfuscated PowerShell as the the logs will show the script as it looks to PowerShell for each phase of execution. There are additional "clean-up" functions which try to help remove additional obfuscation that still exists within the logs and then "filter" functions which remove common-noise that I find in samples.

It's been running for roughly two months without issue in production environments that processes tens of thousands of samples regularly. 

It's basically a souped-up Cuckoo version of this standalone script: http://ropgadget.com/posts/intro_curtain.html


jbremer:
Thanks, good stuff in here! FWIW, I believe @haam3r should have a VMCloak dependency to automatically install the PowerShell extension in newly created VMs.
Other than that I have two comments on this new module:
* it'd be easier to "simplify" show the extracted PowerShell scripts as.. Extracted Artifacts (please see also our 2.0.4/2.0.5 release blogposts)
* would it be possible to create some pure Python code to parse the Windows Event Log? I don't know regarding the feasibility of this, but at least executing PowerShell with a base64-encoded script on the command-line doesn't seem necessary to me.

haam3r:
@jbremer Yes I do, see PR: https://github.com/jbremer/vmcloak/pull/122

Just an idea regarding pure Python code for Event log parsing. @karttoon you could use wevtutil.exe to export the logs in xml and then use the xml library to do the parsing. i.e. something like:

For exporting:
```bash
command = 'wevtutil.exe query-events microsoft-windows-powershell/operational /rd:true /uni:true /e:root /format:xml > {0}'.format(output_path)
```
Convert to json:
```python
import xml.etree.ElementTree as ET

tree = ET.parse(powershell_xml)
        root = tree.getroot()
        dict = etree_to_dict(root)
        with open(analysis_path, 'w') as outfile:
            json.dump(dict, outfile)

def etree_to_dict(t):
        d = {t.tag : map(etree_to_dict, t.getchildren())}
        d.update(('@' + k, v) for k, v in t.attrib.iteritems())
        d['text'] = t.text
        return d
```

jbremer:
Thanks @haam3r. That sounds more like what I was thinking. Simple XML dump from the Guest to Host and some XML parsing on the Host. Thanks for the code snippet.

karttoon:
Sweet, thanks! I actually spent a lot of time trying to find a way to not use PS to get those events and kept coming up blank. I'll work on switching it to wevtutil instead of Get-WinEvent in PS, which will also save some Python filtering since I won't have to actually filter out the Get-WinEvent events anymore.

As for the Extracted Artifacts, I opted for it's own tab since there are effectively three displays for it. The original event, a (potentially) heavily modified de-obfuscated event, and then filtered events. I didn't want to muck with anything existing and try to cram my code into theirs.

Below are two screenshots showing the Original vs Altered (not much obfuscation in this sample but it should illustrate some of what I mentioned above).

original

altered

jbremer:
IMO this is exactly the kind of functionality we'd like to power the "extracted artifacts" tab with. It (the extracted artifacts tab) needs a bit more work from our side, but I believe that's the way forward ;-) If you have ideas on how to essentially merge the tabs, that'd be great.
In the meantime some code around `wevtutil` would be useful. Perhaps an indicator that the PowerShell logging is *missing* would be useful too, so we can instruct users to upgrade their VMs later on.

karttoon:
Converted Aux/Proc scripts to utilize the wevtutil tool and parse the XML. I looked at the Extracted Artifacts page and it appears I'd have to write another function in app.js but I'm not entirely sure. Some of the entries in Curtain, depending on sample, can be quite large with 40-60+ events so I'd expect you wouldn't want them separated. Maybe one "artifact" which represents a PID and within the extracted box it contains the 3 tabs/each individual original event, altered event, and then filtered events.

chudel:
Cool.  I'm not going to step on any toes but let me know if you need some help and I'd be game to give it a shot.  Also - this module was an excellent tutorial into writing/understanding how processing/analysis/web/reporting works inside of cuckoo. Have you considered a kind of hello-world or "empty shell" module that might make it easier for others to extend cuckoo's capabilities?

doomedraven:
@chudel check this https://www.adlice.com/cuckoo-sandbox-customization-v2/

doomedraven:
@karttoon let us know once it will be done, i would like to see it, @jbremer it would be good to reference them somewhere, maybe in documentation as examples, idk

doomedraven:
@karttoon hey, one small thing, i found what sometime(not always) the last file only contains path, but not real log, can you change it to something like, in local with that change works just fine

```
curtLog = os.listdir("%s/curtain/" % self.analysis_path)
curtLog.sort()
for curtain_log in curtLog[::-1]:
            try:
                tree = ET.parse("%s/curtain/%s" % (self.analysis_path, curtain_log))
                root = tree.getroot()
                break
            except Exception as e:
                #raise CuckooProcessingError("Failed opening curtain.log: %s" % e.message)
                pass
```

doomedraven:
so where 2 thigs to bear in mind

1. os.system will be started by explorer, and if something was injected to explorer it start grab behavior of curtain
2. to remove the cmd windows use

```
import subprocess

in def __init__
self.startupinfo = subprocess.STARTUPINFO()
self.startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW

and replace quries
subprocess.call(["C:\\Windows\\System32\\wevtutil.exe", "query-events", "microsoft-windows-powershell/operational", "/rd:true", "/e:root", "/format:xml", "/uni:true"], startupinfo=self.startupinfo,  stdout=open("C:\\curtain.log", "w"))
        

subprocess.call(["C:\\Windows\\System32\\wevtutil.exe", "clear-log", "microsoft-windows-powershell/operational"], startupinfo=self.startupinfo)
```

in this way we make sure it wasn't be tracked in anyway, and no cmd window popups in screens

PR #2063, opened 2018-01-17
Api tasks list sample

kjakan:
Added support for listing tasks for a sample. Requires just a small change to existing tasks list function.

Is this something you would consider including?

Some notes:
* I had trouble getting the unit testing to work, so I have not added a unit test.
* Not sure if the "/v1/" route should also have been there, but assumed that was a reference to Cuckoo 1.*.


jbremer:
Thank you! Multiple people have been asking regarding this feature over time, so thanks for actually developing it.

jbremer:
Ok, manually merged it. Thanks again!

PR #2068, opened 2018-01-18
Removes unused and broken variable

rholloway:
process.calls is an invalid variable in this context, because `_tree.html` template is passing process based on `report.analysis.behavior.processtree` items, which does not actually contain `calls` attribute. `calls` exists and is populated in another piece of the report, but not under `processtree`. 
The pagination still functions and is handled elsewhere, so I don't believe this piece of JSON attribute is actually used anywhere.

Note the error only shows up with DEBUG = True prior to this change, because otherwise the invalid variable will automatically be replaced/handled by django.

### DIFF
--- a/cuckoo/web/templates/analysis/pages/behavior/partials/_process.html
+++ b/cuckoo/web/templates/analysis/pages/behavior/partials/_process.html
@@ -3,7 +3,7 @@
     <p>
       <a href="#">
         {{ process.process_name }}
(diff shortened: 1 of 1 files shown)


jbremer:
Ah, thanks a lot! Didn't realize it wouldn't show up with the `DEBUG` being enabled. Hence couldn't reproduce it locally. Manually merged the commit.

Issue #1619, opened 2017-05-30
Cuckoo Router Not Forwarding Traffic from Guest to Internet


raveydaveygravy:
My Cuckoo setup is working with the exception of providing Internet connectivity to my Win 7 guest.

I can provide Internet connectivity to my guest by using the "old school" iptables rules but not when I use the Cuckoo router.

Guest config:
IP address: 192.168.56.100/24
Gateway: 192.168.56.1
DNS server: 192.168.56.1

Cuckoo routing.conf configuration:

`user01@user01-Workstation:~/.cuckoo/conf$ cat ./routing.conf 
[routing]
# Default network routing mode if none is specified by the user.
# In none mode we don't do any special routing - the VM doesn't have any
# network access (this has been the default actually for quite a while) aside
# from the subnet it exists in.
# In internet mode by default all the VMs will be routed through the network
# interface configured below (the "dirty line").
# And in VPN mode by default the VMs will be routed through the VPN identified
# by the given name of the VPN (as per the VPNs listed in the vpn section).
# Note that just like enabling VPN configuration setting this option to
# anything other than "none" requires one to run utils/rooter.py as root next
# to the Cuckoo instance (as it's required for setting up the routing).
route = none

# Network interface that allows a VM to connect to the entire internet, the
# "dirty line" so to say. Note that, just like with the VPNs, this will allow
# malicious traffic through your network. So think twice before enabling it.
# (For example, to use eth0 as dirty line: "internet = eth0").
internet = eth0

# Routing table name/id for "dirty line" interface. If "dirty line" is
# also default gateway in the system you can leave "main" value. Otherwise add
# new routing table by adding "<id> <name>" line to /etc/iproute2/rt_tables
# (e.g., "200 eth0"). ID and name must be unique across the system (refer to
# /etc/iproute2/rt_tables for existing names and IDs).
rt_table = main

# To route traffic through multiple network interfaces Cuckoo uses
# Policy Routing with separate routing table for each output interface
# (VPN or "dirty line"). If this option is enabled Cuckoo on start will try
# to automatically initialise routing tables by copying routing entries from
# main routing table to the new routing tables. Depending on your network/vpn
# configuration this might not be sufficient. In such case you would need to
# initialise routing tables manually. Note that enabling this option won't
# affect main routing table.
auto_rt = yes

# The drop route basically drops any outgoing network (except for Cuckoo
# traffic) whereas the regular none route still allows a VM to access its own
[...]

PR #2090, opened 2018-01-31
Add iptables rule to fix the source IP on inetsim responses

crahan:
The proposed updates to the inetsim_enable and inetsim_disable functions add an iptables masquerade rule to ensure the correct source IP is put back into redirected packets sent back from the inetsim host. Without this additional rule responses from inetsim will contain the IP address of the inetsim host as the source. Since the IP of the inetsim VM doesn't match the destination IP of the initial request Windows 7 will ignore it. 

The masquerade rule ensures that the source of the response matches the destination of the initial request (i.e. if the Windows host sends a DNS request to 8.8.8.8, the request will be redirected towards the inetsim host, inetsim will send back a DNS response but the source fo this response will be correctly set to 8.8.8.8).

A more detailed discussion can be found at https://github.com/cuckoosandbox/cuckoo/issues/1619#issuecomment-361617430

### DIFF
--- a/cuckoo/apps/rooter.py
+++ b/cuckoo/apps/rooter.py
@@ -197,6 +197,9 @@ def inetsim_enable(ipaddr, inetsim_ip, machinery_iface, resultserver_port):
     dns_forward("-A", ipaddr, inetsim_ip)
     forward_enable(machinery_iface, machinery_iface, ipaddr)
 
+    run(s.iptables, "-t", "nat", "-A", "POSTROUTING", "--source", ipaddr,
+        "-o", machinery_iface, "--destination", inetsim_ip, "-j", "MASQUERADE")
+
     run(s.iptables, "-A", "OUTPUT", "-s", ipaddr, "-j", "DROP")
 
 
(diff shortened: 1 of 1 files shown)


jbremer:
Interesting! Sounds like something we should merge. @doomedraven any feedback on this PR?

doomedraven:
i don't have that rule with kvm, as it looks only vbox related, but i will check it i hope tomorrow and tell if that breaks something or not, but shouldn't anyway, as is pretty simple

crahan:
Just for reference. I actually had to add this on my KVM (and ESXi) Cuckoo Sandbox installs to get Windows (7 and higher) VMs to work correctly with inetsim. Without this iptables rule my analysis VMs would ignore the inetsim responses (even though a Wireshark capture showed they arrived on the Windows VM).

doomedraven:
ok, It looks fine, but just to double check i will verify on my side so we will have double verification :)

doomedraven:
@jbremer sorry for long delay, approved :)

jbremer:
Merged, thanks! Sorry for delay.

crahan:
@jbremer was the merge successful? I'm currently seeing a message about unmerged commits and I also can't see the proposed changes in the most recent rooter.py code.

jbremer:
Yep! Just pushed 'em to upstream Cuckoo.

Issue #2095, opened 2018-02-02
[mitmproxy] transparent mode not allowed


LetMeR00t:
Hello guys,
I've just find a bug in the cuckoo code concerning the use of mitmproxy.
https://github.com/cuckoosandbox/cuckoo/blob/dc1c674b7ca5e0cda8338c82b9e46a2d7300cb83/cuckoo/auxiliary/mitm.py#L83-L86

Due to this part of the code, we are not able to setup mitmproxy as a transparent proxy because it's always setting the parameters "ResultServer IP" and "port" as a proxy.

It should be great, depending on the mode used by mitmproxy, to adapt this behavior :)

jbremer:
Please find my comment at #2102, thanks.

Issue #2093, opened 2018-02-01
[mitmproxy] script is not found


LetMeR00t:
Hello guys,

I'm using the last version of cuckoo (v2-0-5) and I needed to use the last release of mitmproxy (r3-RC2) because the configuration file of mitmproxy was not used.

We found out a bug that I've just detailed here : https://github.com/mitmproxy/mitmproxy/issues/2824

I think the problem is linked to the way on how we give the parameters to mitmproxy here :

https://github.com/cuckoosandbox/cuckoo/blob/dc1c674b7ca5e0cda8338c82b9e46a2d7300cb83/cuckoo/auxiliary/mitm.py#L57-L62

It could be great to resolve this for the next release.

Let me know if you need something else.

mhils:
A fix for this is in https://github.com/cuckoosandbox/cuckoo/pull/2092, sorry for not communicating that on our end. :) We do the rest in mitmproxy with https://github.com/mitmproxy/mitmproxy/pull/2833 so that you can keep one calling convention for any mitmproxy versiom.

LetMeR00t:
FYI, another issue concerning cuckoo/mitmproxy was opened here : https://github.com/cuckoosandbox/cuckoo/issues/2095
It concerns the transparent mode :) It's not the fault of mitmproxy, it's a setting always set on cuckoo's code.

jbremer:
I've merged the #2092 (and will push it in a bit), so this issue can be closed, right? :-)

LetMeR00t:
Yes, I agree :)

PR #2092, opened 2018-02-01
Remove superfluous whitespace on mitmproxy invocation

mhils:
This fixes mitmproxy/mitmproxy#2824.

PR #2102, opened 2018-02-03
[mitmdump] Improved usage thanks to its configuration file

LetMeR00t:
Adding the possibility to configure mitmproxy using his own configuration file.
Moreover, the transparent mode could be used if specified in the configure file.

This is a fix for :
https://github.com/cuckoosandbox/cuckoo/issues/2093
https://github.com/cuckoosandbox/cuckoo/issues/2095

Partially based on the following pull : https://github.com/cuckoosandbox/cuckoo/pull/2092

### DIFF
--- a/tests/test_startup.py
+++ b/tests/test_startup.py
@@ -606,6 +606,7 @@ def test_tmp_permissions_true():
     })
     assert os.path.isdir(temppath())
 
+@pytest.mark.skipif("sys.platform != 'linux2'")
 def test_tmp_permissions_false():
     set_cwd(tempfile.mkdtemp())
     dirpath = tempfile.mkdtemp()
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -76,8 +76,8 @@ class TestCreateFolders:
 
     def test_create_temp(self):
         """Test creation of temporary directory."""
-        dirpath1 = Folders.create_temp("/tmp")
-        dirpath2 = Folders.create_temp("/tmp")
+        dirpath1 = Folders.create_temp()
+        dirpath2 = Folders.create_temp()
         assert os.path.exists(dirpath1)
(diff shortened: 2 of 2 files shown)


jbremer:
@LetMeR00t @mhils this PR looks good, except for the part where a Python 3-only version is required. Is there a way to somehow work around that for now?
As you're aware we'll be working on integrating some of the mitmproxy `0.18.2` stuff in Cuckoo until Cuckoo is Python 3-compatible (which will certainly be a while still).

LetMeR00t:
Hi @jbremer , 
Did you mean mitmproxy v3 release ? Because it's what I wrote as a comment in the code and I don't say anything according to Python 3. Am I wrong ?

jbremer:
The last mitmproxy version that supports Python 2 is `0.18.3` is what I meant and I'm interested in using that exact version for some other purposes, so installing mitmproxy 3 which is Python 3-only would hinder that (I can only assume).

LetMeR00t:
Do you mean that we can't push the "configuration file parameter" yet ?
Did you have some trouble to use the mitmproxy > v0.18.3 ? it's used by your httpreplay if I understand but it seems to be not so complex and adapt, am I wrong ?

jbremer:
As mentioned Cuckoo is Python 2 currently, so yes, that'd make some things that are coming up more difficult, I believe.

LetMeR00t:
Ok well, I understand but I recommand to remove these lines :
https://github.com/cuckoosandbox/cuckoo/blob/59d32361c1636b2b3802a1746f480a7768f6384f/cuckoo/auxiliary/mitm.py#L83-
In my case, i'm using mitmproxy v3-RC2 in transparent mode and this part of code is not allowing me that. These parameters should be set manually by the user in the VM web browser, don't you think ?

Issue #2057, opened 2018-01-14
When adding a new reporting model the config section is not auto-detected.


KasperskyZiv:
As far as i could see, and correct me if i'm wrong, but my new section in the reporting config wasn't detected automatically. I had to go over some of the config code before i added a new "schema" to **Config.configuration** field like this:
```
"reporting": {
            "myModel": {
                "field": Boolean(False),
                "another_field": Path(
                        exists=True, writable=False, readable=True
                ),
                "url": String(sanitize=True),
                "username": String(),
                "password": String()
    }
}
```
Is this how i'm to make it work?
If so why is it not updated in the manual..

Thanks!
ziv

doomedraven:
is to verify what a user has correct config values

krisleonard-mcafee:
Would it be possible for Cuckoo to add a required configuration definition method to the abstract Report object? This would allow developers adding new reporting modules to define the configuration requirements in their own code and not require modification of Cuckoo's config.py.

jbremer:
Yes, that'd make sense and we've considered in the past already, but simply didn't have the time to implement it yet. Feel free to do a PR ;-)

jbremer:
Closing issue as there's a PR now. Merging the PR will require some more thoughts, though. To be done in the future.

Issue #1580, opened 2017-05-19
[Customization] Config section not found! 


Tigzy:
Hello, 
I used to work on cuckoo modules to customize my scan, I will take as example a module that submits the dropped files to my webserver as example:

So I put my module "dropped_upload.py" into /usr/lib/pythong2.7/dist-packages/cuckoo/processing (though I think it would be great to have the ability to put them into CWD/modules...)

I also add the section into processing.conf
```
[dropped_upload]
enabled = yes
```

When I run cuckoo I have weird results (it used to work until 2.0 RC1):
```
Configuration error: Config section processing:dropped_upload not found!
```
Imported "processing" modules:
    |-- DroppedUpload
```

When I run an analysis, unfortunately the module isn't called:
```
DEBUG: Processing module dropped_upload not found in configuration file
```

What am I missing?

jbremer:
We have much more strict configuration checks now, see also `cuckoo/common/config.py`, as there have been quite some people with incorrectly filled out configuration settings resulting in impossible-to-debug user setup issues - this entire issue appears to be pretty much solved now (as no new github issues are created around such topics).
Naturally this also means that you'll have to add additional entries in `config.py` for your plugin. The additional factor that this somewhat suggests that contributing back your modules is easier than to keep a local setup up-to-date, well, is something to be said for as a community project ;-)
You aren't the first to request, e.g., `$CWD/modules/`, though, and I've thought about supporting something like that - I might get to implementing it in a bit.

Tigzy:
This means I have to edit config.py to add my own modules? Sounds a bit uncommon and against the modular way to do...

jbremer:
At the moment, yes, for reasons I explained above. We'll probably improve this once again in the future, though.

doomedraven:
@Tigzy well is modular, but in safe way, you just need to add an entry to verify the config. and if that get merged it will be inside ;)

Tigzy:
ok thanks, it's working after I've added the proper fields. I hope we find a better way in the future

rholloway:
Just encountered this issue as well. Having to modify config.py directly is not ideal at all in my opinion, not to mention the docs need to be updated if this is the new way forward. Only way I figured it out was by reading through all the source code.

Tigzy:
Is this still working? I have the problem again after upgrade

PR #2112, opened 2018-02-06
Allow Reporting, Auxiliary, and Processing plugins to specify their configuration instead of having to modifying Config.configuration in config.py

krisleonard-mcafee:
Changes to allow Reporting, Auxiliary, and Processing plugin developers to specify their configuration definitions instead of having to modify Config.configuration in config.py. Please see https://github.com/cuckoosandbox/cuckoo/issues/2057 and https://github.com/cuckoosandbox/cuckoo/issues/1580 for background on the issue resolved with this change. 

This is just a initial idea to see if the Cuckoo team thinks this is the right track. I just wanted to get the ball rolling with the Cuckoo team before making more changes. The items that I think need to be added still are:
- Validation of the data returned by get_config() before adding it to "Config.configuration".
- Unit tests in test_startup.py, maybe test_config.py, and any where else the Cuckoo team would recommend. 

Thanks for any feedback!

### DIFF
--- a/cuckoo/common/abstracts.py
+++ b/cuckoo/common/abstracts.py
@@ -39,6 +39,10 @@ class Auxiliary(object):
         self.options = None
 
     @classmethod
+    def get_config(cls):
+        pass
+
+    @classmethod
     def init_once(cls):
         pass
 
@@ -610,6 +614,10 @@ class Processing(object):
         self.machine = None
         self.options = None
         self.results = {}
+
+    @classmethod
+    def get_config(cls):
+        pass
 
     @classmethod
     def init_once(cls):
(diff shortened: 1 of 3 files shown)


cuckoo/common/abstracts.py
```diff
@@ -38,6 +38,10 @@ def __init__(self):
         self.guest_manager = None
         self.options = None
 
+    @classmethod
```
krisleonard-mcafee: I went with using a "classmethod" so that the new update_config method would not have to create an instance of all the various plugins. I do not imagine plugin developers will make use of the fact that it is a "classmethod" but this seemed better than creating a new instance of all the plugins in update_config(). Also the change does not require plugin developers to implement "get_config" and they can continue to modify config.py if they wish too.

PR #2194, opened 2018-03-27
Ensure socket is closed after connection test

jcppkkk:
Ensure socket is closed after connection test against guest

While doing mass analysis sometimes sandbox VM hangs. 
If that happens, guest initialization will hangs and the guest stops without any log or exception.
I try to ensure every works are well used during mass analysis.

Doing analysis over 10,000 samples with 300+ guest vm after this change.
No guest hangs after this change.

### DIFF
--- a/cuckoo/core/guest.py
+++ b/cuckoo/core/guest.py
@@ -8,6 +8,8 @@ import io
 import json
 import logging
 import os
+from contextlib import closing
+
 import requests
 import socket
 import time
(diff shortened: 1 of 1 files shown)


jcppkkk:
### socket.create_connection() correctly closes the socket upon exception, so it would appear the socket is always closed in the existing code

It modified because it actually hangs on massive runs, i have no idea why the originally code hangs, but the modified code works (not hangs at this point anymore).

I read the doc again and:
https://docs.python.org/2/library/socket.html#socket.create_connection
> socket.create_connection(address[, timeout[, source_address]])
> Passing the optional timeout parameter will set the timeout on the socket instance before attempting to connect. If no timeout is supplied, the global default timeout setting returned by getdefaulttimeout() is used.

https://docs.python.org/2/library/socket.html#socket.getdefaulttimeout
> Return the default timeout in seconds (float) for new socket objects. A value of None indicates that new socket objects have no timeout. When the socket module is first imported, the default is None.

If `socket.create_connection()` is not going to change, then i suggest set a timeout for it ⏳ 

### We believe that re-doing a GET/POST request isn't necessary in this context and instead merely providing the request timeout is necessary.

It's true, this is intent to be a workaround of other problems like `self.get("/pinning")` action sometimes return 500 error. However this can be separated to a new issue.

PR #2220, opened 2018-04-05
add support to provide banned sids when checking suricata

doomedraven:

add support to provide banned sids when checking suricata
add more power to suricata checking

### DIFF
--- a/cuckoo/common/abstracts.py
+++ b/cuckoo/common/abstracts.py
@@ -1128,13 +1128,14 @@ class Signature(object):
                                  regex=regex,
                                  all=all)
 
-    def check_suricata_alerts(self, pattern):
+    def check_suricata_alerts(self, pattern, banned_sids=[]):
         """Check for pattern in Suricata alert signature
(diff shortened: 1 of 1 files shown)


rholloway:
@doomedraven is this available to be set via the config file somewhere? This could be handy for some of the noisier signatures if someone is using default config and rulesets (otherwise they do need to manage and exclude rules separately with other tooling, doable but a simple config for certain sid's could be handy)

doomedraven:
@rholloway  they already works on it :)

lillypad:
@doomedraven there is already a file to store whitelisted domains for whitelisted suricata signature ids they could be kept int he same manor. You would need to add a function to this to handle reading a file of sid numbers. To go with this. 

Logically it would make sense if you were to do this to add suricata_sids to the whitelist folder here: https://github.com/cuckoosandbox/cuckoo/tree/master/cuckoo/data/whitelist

doomedraven:
Normally i have this problem with ips, shared hacked routers as proxis

PR #2252, opened 2018-04-28
A per-host pid mask

rx13:
Added the option for a per-host PID mask.
To permit handling of multiple environments in the same cuckoo instance where pre-attack PIDs may vary.
Added changes over the most recent master branch and validated with multiple VMs.

##### Other notes
Implementation was done to match style/format of preexisting pid_generic functionality, and to concatenate. Allowing for pid_generic to be a source of shared excludes where someone may have already implemented it, and per-host specific excludes when needed.


rx13:
I also had that concern but the functionality already exists on a global setting today, and thus far reports before and after setting masks have seen no difference in accuracy that I can tell.

PR #2365, opened 2018-07-11
avoid SSL/TLS problem with recent vSphere 6.7 release

### LINKED ISSUE #2318: unverified_ssl=yes breaks on recent vCenter versions that doesn't support TLSv1
##### The log, error, files etc can be found at:
in the vsphere.py code these lines manage unverified ssl setting: 
        # Workaround for PEP-0476 issues in recent Python versions
        if self.options.vsphere.unverified_ssl:
            sslContext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
            sslContext.verify_mode = ssl.CERT_NONE
            self.connect_opts["sslContext"] = sslContext
            log.warn("Turning off SSL certificate verification!")

In this case, forcing ssl protocol to TLSv1 in not a good idea because probably vSphere 6.7 does not support this version (probably because of POODLE attack: https://www.globalsign.com/en/blog/poodle-vulnerability-expands-beyond-sslv3-to-tls/). Solved it on my installation changing the second line to:  sslContext = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2), but it does not seem to be the right way to solve this. In https://www.python.org/dev/peps/pep-0476/ it recommends using: 
 . context = ssl._create_unverified_context() 

update: changed this part of the code to:
     # Workaround for PEP-0476 issues in recent Python versions
        if self.options.vsphere.unverified_ssl:
            sslContext = ssl._create_unverified_context()
            self.connect_opts["sslContext"] = sslContext
            log.warn("Turning off SSL certificate verification!")

kalos:
TLSv1 is unsupported in recent vSphere.
Using '_create_unverified_context' for create correct context.

This resolve #2318

### DIFF
--- a/cuckoo/machinery/vsphere.py
+++ b/cuckoo/machinery/vsphere.py
@@ -96,8 +96,7 @@ class vSphere(Machinery):
 
         # Workaround for PEP-0476 issues in recent Python versions
         if self.options.vsphere.unverified_ssl:
-            sslContext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
-            sslContext.verify_mode = ssl.CERT_NONE
+            sslContext = ssl._create_unverified_context()
             self.connect_opts["sslContext"] = sslContext
             log.warn("Turning off SSL certificate verification!")
 


jbremer:

PR #2454, opened 2018-08-28
Add missing .vbs extension if needed in vbs package

FernandoDoming:

Automatically add .vbs extension in the VBS package if the files does not have it because wscript.exe requires the extension to be present.


jbremer:
Manually merged, thanks!

PR #2486, opened 2018-09-19
Add a custom kvm libvirt uri dsn cwd configuration option in conf/kvm.conf machinery option. Update the …

wroersma:
Add environment variable option for kvm machinery option
Make it easier to override the kvm dsn option instead of having to override the entire python file.
I have tested my changes with tcp and ssh type of connection strings set with the LIBVIRT_URI option.


doomedraven:
do you know what you can set that in qemu config or as env var?

PR #2513, opened 2018-10-05
Allow specifying list of IRMA probes to use

haam3r:
Add a new optional config setting for specifying a list of probes to the IRMA processing module.

This introduces the possibility of specifying which probes to use in IRMA when submitting from Cuckoo i.e. for example exclude VirusTotal when it's already being done by Cuckoo itself 

- That the probe list is respected and used.
- If no probes option is specified Cuckoo will default to using all available probes
- That nothing breaks with result retrieval or anywhere else


jbremer:
Manually merged, thanks!

PR #2518, opened 2018-10-10
Sysmon processing / auxiliary module

FernandoDoming:
Inspired by: https://github.com/cuckoosandbox/cuckoo/pull/2042 a somewhat similar module for sysmon.

A Sysmon harvesting / processing module for Cuckoo. The guest needs to have Sysmon installed & configured for this to work.

Get Sysmon information into Cuckoo.

Deployed locally

### DIFF
--- a/cuckoo/private/cwd/conf/auxiliary.conf
+++ b/cuckoo/private/cwd/conf/auxiliary.conf
@@ -80,3 +80,6 @@ timeout = {{ auxiliary.services.timeout }}
 [reboot]
 # This auxiliary module should be enabled for reboot analysis support.
 enabled = {{ auxiliary.reboot.enabled }}
+
+[sysmon]
+enabled = {{ auxiliary.sysmon.enabled }}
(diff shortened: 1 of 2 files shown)


wroersma:
Hey one thing I noticed is that you don't have xmltodict added to be installed as well as the unit tests for this change in config file management.

cccs-kevin:
Note that this PR does not work out-of-the-box. As of v2.0.7+, the following changes are required on top of the file changes indicated in this PR:
- [ ] It is required to add `"sysmon"` to the `RESULT_UPLOADABLE` list in `cuckoo/core/resultserver.py`
- [ ] The `xmltodict`python library needs to be installed (`pip install xmltodict`) as per https://github.com/cuckoosandbox/cuckoo/pull/2518#issuecomment-467151174
- [ ] Per https://github.com/cuckoosandbox/cuckoo/pull/2042#issuecomment-437306411, change `os.system` calls to `subprocess.call` to avoid potential injection, and use `self.startupinfo` as indicated in comment

In my setup, sysmon.exe was not running in my VM by default on startup so the following changes were required:
- [ ] Add sysmon.exe from https://download.sysinternals.com/files/Sysmon.zip to `$CWD/analyzer/windows/bin/`
- [ ] Add a start() method to the `sysmon.py` auxiliary module that uses sysmon.exe to make the following call: `subprocess.call([os.path.join(self.analyzer.path, "bin", "sysmon.exe"), "-accepteula", "-i"], startupinfo=self.startupinfo)`

Hopefully this is useful to someone...

PR #2519, opened 2018-10-10
Add configuration option to kvm.conf for custom dsn option use.

wroersma:
What I have added/changed is:
Add config option for kvm machinery option in kvm.conf

The goal of my change is:
Make it easier to override the kvm dsn option instead of having to override the entire python file.

What I have tested about my change is:
I have tested my changes with tcp and ssh type of connection strings set with the dsn option in kvm.conf.

### DIFF
--- a/cuckoo/common/config.py
+++ b/cuckoo/common/config.py
@@ -391,6 +391,7 @@ class Config(object):
         },
         "kvm": {
             "kvm": {
+                "dsn": String("qemu:///system", required=False),
                 "interface": String("virbr0"),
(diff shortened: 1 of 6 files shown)


doomedraven:
why just not to read the config with dsn?

wroersma:
Read which config with dsn? It wasn't a config option to set a custom dsn option for kvm before. I might be confused on what you're asking here.

doomedraven:
check `/etc/libvirt/libvirt.conf`

PR #2538, opened 2018-10-23
Remove false positive PEiD Armadillo signatures from peutils UserDB

wesinator:

Removed the PEiD Armadillo v1.71 signature from peutils signature db

The PEiD Armadillo v1.71 signature is incorrect, the sequence matches on any binaries compiled with certain versions of MSVC, causing false positive packer detections.

Ref https://github.com/brad-accuvant/cuckoo-modified/commit/defb7b42120583c0d56efa68698ec3a625e106bf

None


wesinator:
PEiD "Armadillo v1.xx - v2.xx" is also a false positive 

https://www.zscaler.com/blogs/research/your-windows-8-packed

PR #2580, opened 2018-12-13
TAG + Minimal score + Publication settings

Tux-Panik:
Adding the following features:
- Tag capabilities to MISP
- Minimal score to avoid irrelevant publications
- Publication level reviewed to limit them to "Your Organization"

To be able to control and/or restrict the publications to MISP

The changes passed my tests.
Publication works.

### DIFF
--- a/cuckoo/reporting/misp.py
+++ b/cuckoo/reporting/misp.py
@@ -5,120 +5,135 @@
 import os.path
 import shlex
 import warnings
+import json
 
 from cuckoo.common.abstracts import Report
 from cuckoo.common.exceptions import CuckooProcessingError
 
 class MISP(Report):
-    """Enrich MISP with Cuckoo results."""
+		"""Enrich MISP with Cuckoo results."""
 
-    def sample_hashes(self, results, event):
-        """For now only reports hash of the analyzed file, not of the dropped
-        files, as we may have hundreds or even thousands of dropped files, and
-        the misp.add_hashes() method doesn't accept multiple arguments yet."""
-        if results.get("target", {}).get("file", {}):
-            f = results["target"]["file"]
-            self.misp.add_hashes(
-                event,
-                category="Payload delivery",
-                filename=f["name"],
-                md5=f["md5"],
-                sha1=f["sha1"],
-                sha256=f["sha256"],
-                comment="File submitted to Cuckoo",
-            )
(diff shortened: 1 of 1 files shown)


doomedraven:
you don't follow the collaboration guide from documentation, no tabs
+ with open('.cuckoo/storage/analyses/' + str(uid) + '/reports/report.json', 'r') as json_file:
can be easilly replaced by `cwd`

SparkyNZL:
awesome :) I to have been working on this, but now i dont need to :) I was
also adding to the "Seen" when it detects a hit :)

doomedraven:
still need replace tabs to spaces

Tux-Panik:
What a grumpy!!!
Change ongoing :-)

doomedraven:
ROFL you want your PR die like the rest of PRs with more reason? :D
https://cuckoo.readthedocs.io/en/latest/development/code_style/

doomedraven:
anyway that still not respect code style

doomedraven:
as per my idk as i don't use misp, but if you respect code style it make it much more possible to get merged

Tux-Panik:
I understand this point of view, for sure.
Often the same problem: I'm not a dev guy, so I'm lost with your (dev guys) specifications, prerequisites and so on. From my perspective, those kinds of stuff don't seem so important... but I'm probably wrong and you've more experience than I!!!
I'll try to change what I can and what I understand in your remarks.
Thanks for your time and support,
Regards,
Julien

doomedraven:
it wasn't important for me neither before, but im already a cuckoo user for 8y, so imagine if there would be so different to read/maintain/fix code , is why they put that requirement ;)

basically everything is here https://cuckoo.readthedocs.io/en/latest/development/code_style/

rkokkelk:
@Tux-Panik, great to see some work being done on misp.py. However, I think your PR still has some issues, mainly:

- TAG & publication is currently hardcoded, please make that a configuration option
- You use CWD as a (undeclared) variable which will fail. doomedraven meant you should use the cwd function, declared [here](https://github.com/cuckoosandbox/cuckoo/blob/master/cuckoo/misc.py#L43), for example usage see [here](https://github.com/cuckoosandbox/cuckoo/blob/master/cuckoo/reporting/elasticsearch.py#L55)

I'm quite interested in this functionality. So if you need any dev help, just let me know!

PR #2617, opened 2019-01-21
IP forwarding is required always

utkonos:
Just make it clear in the docs that ip forwarding is required no matter global or per-analysis.


jbremer:
This has been included in the 2.0.7 release, thanks!

PR #2637, opened 2019-02-04
VT query for dropped PE32/MS-DOS binaries ,not usefully functioning

Tatsuya-hasegawa:
Hi cuckoo team

Nice to see you again.
I may have found a strange code in virustotal.py

The VT result of dropped PE binaries was saved to row['virustotal'].
The valuable "row" is a local variable in a for loop. 
Thus, it is not reduced to "results".
The code was just the function only uploading the dropped files to VT for specific user which they set "scan = yes" on their configuration.

On the other hands, malware sometimes drops a lot of PE32/MS-DOS binaries.
It will cause to waste the api license in many cases.

If all of the VT results should be saved at default automatically, 
I think we had better reduce the data stored in "results".

Kind Regards,
Tatsuya

### DIFF
--- a/cuckoo/apps/api.py
+++ b/cuckoo/apps/api.py
@@ -1,5 +1,5 @@
 # Copyright (C) 2012-2013 Claudio Guarnieri.
-# Copyright (C) 2014-2018 Cuckoo Foundation.
+# Copyright (C) 2014-2019 Cuckoo Foundation.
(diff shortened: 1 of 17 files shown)


Tatsuya-hasegawa:
I have made  a mistake that I had merged my forked master origin in spite of this PR from my master origin. Thus I have closed this PR once.

PR #2700, opened 2019-03-23
Ability to whitelist ips

RicoVZ:
Adding IPs to $CWD/whitelist/ip.txt will cause them to
be ignored during network processing.


jbremer:
Thanks, manually merged.

PR #2705, opened 2019-04-01
Fix/recursive plugins

psrok1:
- Added support for nesting plugin packages
- Package walker uses `pkgutil.iter_modules` instead of `os.listdir`

Sometimes there is a need to provide signatures for few versions of the same malware family. These signatures are usually using common code base, so keeping them in separate subpackage seems to be the most elegant way.

Example:
```
signatures
├── emotet
│   ├── __init__.py
│   ├── emotet_v4.py
│   ├── emotet_v5.py
│   ├── libs
│   │   └── ... 
```

Unfortunately, in current version - only flat structure of plugins is supported.

- Necessary unit tests included in this PR

### DIFF
--- a/cuckoo/core/plugins.py
+++ b/cuckoo/core/plugins.py
@@ -7,6 +7,7 @@ import os
 import importlib
 import inspect
 import logging
+import pkgutil
 
 import cuckoo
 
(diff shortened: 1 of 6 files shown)


jbremer:
Thanks!

PR #2723, opened 2019-05-07
Initial support for Proxmox Virtual Environment

michaelweiser:
Add a machinery module allowing to access Proxmox-based virtual machines via
its REST API.

Based on #1831. Does not, however, rely on the Proxmox VM ID to find the VM and
instead uses the configured label, avoiding the need to extend the Cuckoo
database schema with this machinery specific information.

Adds the proxmoxer python module for REST API access as a
dependency.

Add a proxmoxer machinery module complete with configuration template and updated python module dependency.

Add out-of-the-box-support for Proxmox Virtual Environment in Cuckoo.

- successfully starts and runs analyses using PVE in a test environment
- correctly reports on the various error conditions such as wrong credentials, unreachable host or timeouts expiring
- correctly preconfigures the configuration template upon install
- installs the proxmoxer python module dependency in the correct version when installed using setup.py in a fresh virtualenv
- successfully ran the test suite and verified that it does not introduce (new) test failures


michaelweiser:
Travis failure `RuntimeError: WeasyPrint does not support Python 2.x anymore. Please use Python 3 or install an older version of WeasyPrint.` seems unrelated to this change and might require changes to the pip call in `.travis.yml` to `pip install psycopg2 mysql-python m2crypto==0.24.0 weasyprint==0.36" analogous to `setup.py` to make travis work again.

michaelweiser:
Test suite failure in `tests/test_signatures.py`: `E   AttributeError: 'NoneType' object has no attribute 'endswith'` is also present on master and not introduced by this change.

RicoVZ:
Hi @michaelweiser,

Thanks for making the PR. Nice idea and it looks good on a first glance. 
I will put it on my list to test. 

If anyone else has tested this/is willing to test this with Proxmox, please share your experiences. :smile:

githule:
Hello, 
I am using proxmox in a custom install that is not ready at all for redistribution (I disabled most configuration checks from the official code and used workarounds for some bugs inherent to my implementation). 

Today I have no time and no testing platform available for **this PR** but I can bring my improvements/suggestions if possible.

At the moment, I have functional cuckoo with full memory dumps and working to merge my fork with the last Cuckoo 2.0.7. 
I didn't implemented cuckoo rooter and guacd (as the interaction with VM can be obtained directly in the Proxmox hypervisor).
I am interested by an official support and may contribute as soon as I have some free time.

michaelweiser:
My selection of proxmoxer was based purely on the fact that it seemed more actively maintained than others at the time (an impression confirmed by a new release on PyPI on 2020-01-13, the last pyproxmox release being 2013-04-18). It also claims at least some Python 3 support (3.4), the rationale being not to add another migration blocker with this new feature even though cuckoo may be python2 only currently.

Regarding rooter and memdumps: I have simply had no reason yet to add any proxmox-specific bits there. I am however happy to test and possibly integrate anything you have.

What I did run into was the fact that the Linux virtual bridge interfaces used by Proxmox work like a switch and the Cuckoo VM running `tcpdump` will not be able to see all the traffic on the analysis network but only the traffic directed at itself and broadcasts. This is particularly important if you're running inetsim but in a separate VM and not in the cuckoo VM. I was able to work around that by disabling MAC address learning on the whole bridge using `brctl setageing vmbrX 0` or, alternatively,  `bridge link set dev tunX learning off` on the individual inetsim and analyses VM network interfaces, which basically turns the bridge into a hub as far as those ports are concerned. In reflex I dismissed this as a Proxmox setup detail. But if there's any way the machinery module could take care of this, I would certainly like to implement that.

I am part of a team who is actively using and improving this as we go along and would very much like to see it included. As you can see from the massive race-condition I implemented initially, it presents a bit of a learning curve for me. So any advice on how to improve the module and more natively integrate it with Cuckoo would be highly appreciated.

vinceplayer:
memory dumps doesn't work how can is possible to implemented it?

githule:
> memory dumps doesn't work how can is possible to implemented it?

You can run it via the Qemu monitor console. Dont know if your proxmoxer api module handle this.
I did it with the proxmox API that can handle it via : 
`POST /api2/json/nodes/{node}/qemu/{vmid}/monitor`

Run 
`qm monitor <vmid>`

The monitor command I used is 
`dump-guest-memory <target_path>`

Sorry not having much time to spend on this and doesn't have a test plateform for it.

Would be cool to see cuckoo officially integrating proxmox

PR #2738, opened 2019-05-22
block public file download

amadisson:
What I have added/changed is: Create middleware that blocks file download if ipv4 or ipv6 address not in allowed subnets/ip list. 
The goal of my change is: To block public download of malware samples and dropped files so only admins can download and users can't download other users' files. Not to allow users to use Cuckoo to distribute malware.  
What I have tested about my change is:  file download works if in list or no list in local settings, file download blocked if not in list.

### DIFF
--- a/cuckoo/web/utils.py
+++ b/cuckoo/web/utils.py
@@ -132,3 +132,14 @@ def normalize_task(task):
             os.path.basename(task["target"])
         )
     return task
+
+def get_client_ip(request):
+    try:
+        x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR')
+        if x_forwarded_for:
+            ip = x_forwarded_for.split(',')[0]
+        else:
+            ip = request.META.get('REMOTE_ADDR')
+        return ip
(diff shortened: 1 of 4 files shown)


jbremer:

cuckoo/web/utils.py
```diff
@@ -132,3 +132,14 @@ def normalize_task(task):
             os.path.basename(task["target"])
         )
     return task
+
+def get_client_ip(request):
+    try:
+        x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR')
```
Inndy: `X-Forwarded-For` can be spoofed easily.

ref: https://portswigger.net/kb/issues/00400110_spoofable-client-ip-address

RicoVZ: We agree, this part was edited out and merged manually after that, without that header. :slightly_smiling_face:

https://github.com/cuckoosandbox/cuckoo/blob/master/cuckoo/web/web/middle.py#L36

PR #2813, opened 2019-07-25
Add new option to send dropped files to MISP

zaphodef:
This PR introduces the possibility to push the dropped files into the created MISP event. Note that we need `pymisp==2.4.111.2` which fixes a bug, see MISP/PyMISP#321. This version does not break any already in use MISP feature. This new feature has been tested on a local setup (MISP/cuckoo).

Also note we can only use deprecated methods from `pymisp`, as the new recommended methods only work on Python >= 3.6 (see [the `pymisp` documentation](https://pymisp.readthedocs.io/modules.html#pymispexpanded-python-3-6-only))


zaphodef:
(gonna fusion 3 PRs)

PR #2842, opened 2019-08-23
fix for tcpdump user permission check

noideal:

tcpdump user check against file permissions

check user permission before using TCPDUMP

Tested on Ubuntu. This will possibly have problems on windows systems, depending on how it handles the backslash in the file path. I have no way to test this at the moment.


doomedraven:
== should be `is True`

PR #2869, opened 2019-09-12
REST API doc: Use curl parameter -u instead of -H

ASMag:
Curl parameter -H can be used for specifying an extra HTTP header. The right parameter to specify user and password (or a token) is -u. If user or password contains non ASCII characters curl will encode them. If you use parameter -H there is no encoding and your authentication respectively request will fail.

Documentation how to use RestAPI.

To keep others from falling into the same trap I did.

I did some RestAPI requests to my cuckoo instance.

### DIFF
--- a/docs/book/usage/api.rst
+++ b/docs/book/usage/api.rst
@@ -199,7 +199,7 @@ Adds a file to the list of pending tasks. Returns the ID of the newly created ta
 
 **Example request**::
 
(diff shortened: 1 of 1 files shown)


ASMag:
I don't understand why the Travis CI build failed. My pull request only addresses the doc of cuckoo project. Where is the problem?

Issue #2823, opened 2019-07-31
Can cuckoo support Linux guest well?


myhsu:
Thanks for creating an issue! But first: did you read our community guidelines?
https://cuckoo.sh/docs/introduction/community.html

##### My issue is:
1. I always can't get the behavior analysis result after cuckoo analyzed Linux sample. 
2. It seems to use Windows API to analyze Linux memory dump. 
##### My Cuckoo version and operating system are:
1. Cuckoo version: 2.0.7
2. Host OS: Ubuntu 18.04
3. Guest OS: Ubuntu 18.04
##### The log, error, files etc can be found at:
1. 
===
2019-07-29 11:07:24,765 [cuckoo.processing.behavior] WARNING: Analysis results folder does not contain any behavior log files.
===
2. Logs
===
2019-07-29 10:27:51,506 [cuckoo.core.plugins] ERROR: Failed to run the processing module "Memory" for task #27:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/cuckoo/core/plugins.py", line 246, in process
    data = current.run()
  File "/usr/local/lib/python2.7/dist-packages/cuckoo/processing/memory.py", line 1118, in run
    return VolatilityManager(self.memory_path, osprofile).run()
  File "/usr/local/lib/python2.7/dist-packages/cuckoo/processing/memory.py", line 1039, in run
    results[plugin_name] = getattr(self.vol, plugin_name)()
  File "/usr/local/lib/python2.7/dist-packages/cuckoo/processing/memory.py", line 172, in pslist
    for process in command.calculate():
  File "/usr/local/lib/python2.7/dist-packages/volatility-2.6.1-py2.7.egg/volatility/win32/tasks.py", line 88, in pslist
    for p in get_kdbg(addr_space).processes():
  File "/usr/local/lib/python2.7/dist-packages/volatility-2.6.1-py2.7.egg/volatility/win32/tasks.py", line 50, in get_kdbg
    if obj.VolMagic(addr_space).KPCR.value:
  File "/usr/local/lib/python2.7/dist-packages/volatility-2.6.1-py2.7.egg/volatility/obj.py", line 751, in __getattr__
    return self.m(attr)
  File "/usr/local/lib/python2.7/dist-packages/volatility-2.6.1-py2.7.egg/volatility/obj.py", line 733, in m
    raise AttributeError("Struct {0} has no member {1}".format(self.obj_name, attr))
[...]

PR #2897, opened 2019-11-02
Fix bug: "Netlog client requested banned path: logs/all.stap"

TripleS0:
change line 49 of resultserver.py from "if dir_part not in RESULT_UPLOADABLE:" to "if dir_part not in RESULT_DIRECTORIES:". 
Fix bug: resultserver always bans to receive all.stap from guest Linux. This issue was also found in #2823. 
The code works well for the sample analysis in guest Linux.


junquera:
As I commented in #2823 if this was implemented, malware could overwrite the logs (its easier to overwrite files in `logs` path whose name is derived from *pid*, than arbitrary names elsewhere). I think it should be fixed with something like #2986

PR #2938, opened 2020-01-08
Cryptography

utkonos:

This PR removes the M2Crypto library certificate parsing code and the swig dependency for that library. The new code uses python cryptography: https://github.com/pyca/cryptography

The display code in the web component is changed to handle the data fields.

The documentation is also updated.

Use the modern cryptography library that does not have a swig dependency.

I have been using this patch set for about 8 months in production.

### DIFF
--- a/cuckoo/processing/static.py
+++ b/cuckoo/processing/static.py
@@ -3,6 +3,8 @@
 # This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
 # See the file 'docs/LICENSE' for copying permission.
 
+import base64
+import binascii
 import bs4
 import ctypes
 import datetime
@@ -20,18 +22,17 @@ import struct
 import zipfile
 import zlib
 
-try:
-    import M2Crypto
-    HAVE_MCRYPTO = True
-except ImportError:
-    HAVE_MCRYPTO = False
-
 from cuckoo.common.abstracts import Processing
 from cuckoo.common.objects import Archive, File
 from cuckoo.common.structures import LnkHeader, LnkEntry
 from cuckoo.common.utils import convert_to_printable, to_unicode, jsbeautify
 from cuckoo.core.extract import ExtractManager
 from cuckoo.misc import cwd, dispatch
+
+import cryptography
+from cryptography.hazmat.backends.openssl.backend import backend
+from cryptography.hazmat.backends.openssl import x509
+from cryptography.hazmat.primitives import hashes
 
 from elftools.common.exceptions import ELFError
 from elftools.elf.constants import E_FLAGS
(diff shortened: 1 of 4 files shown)


doomedraven:
tested, works just fine

utkonos:
> tested, works just fine

Awesome. I'm working on the additional fields that you suggested. Should have those added to this PR shortly.

RicoVZ:
Thanks @utkonos :slightly_smiling_face: Will merge when making a new version.

cuckoo/processing/static.py
```diff
@@ -250,45 +251,61 @@ def _get_signature(self):
...
-                "md5": "%032x" % int(cert.get_fingerprint("md5"), 16),
-            })
-
-            if subject.GN and subject.SN:
-                ret[-1]["full_name"] = "%s %s" % (subject.GN, subject.SN)
-            elif subject.GN:
-                ret[-1]["full_name"] = subject.GN
-            elif subject.SN:
-                ret[-1]["full_name"] = subject.SN
+        signers = backend._lib.PKCS7_get0_signers(pkcs7_obj, backend._ffi.NULL, 0)
```
King-Konsto: This does not work with enveloped digital certificates. Even though you are running with a for loop, there will always be only one certificate as a result from PKCS7_get0_signers. I haven' get it running with crypotgraphy, but I have with pyopenssl.
`from OpenSSL import crypto

    certs = crypto._ffi.NULL
    pkcs7 = crypto.load_pkcs7_data(
        crypto.FILETYPE_ASN1,
        extract_signature(pe)
    )

    if pkcs7.type_is_signed():
        certs = pkcs7._pkcs7.d.sign.cert
    elif pkcs7.type_is_signedAndEnveloped():
        certs = pkcs7._pkcs7.d.signed_and_enveloped.cert

    pycerts = []
    for i in range(crypto._lib.sk_X509_num(certs)):
        x509 = crypto._lib.X509_dup(crypto._lib.sk_X509_value(certs, i))
        pycert = crypto.X509._from_raw_x509_ptr(x509)
        pycerts.append(pycert)
    if pycerts:
        return pycerts`

King-Konsto:
This does not work with enveloped digital signatures (see comment), old code does also not work if there is more than one digital signatures appending to a pefile.

cuckoo/processing/static.py
utkonos: I had a thread on the mailing list for the cryptography library about multiple certs. Last I checked, it was being worked on. I'll take a look at fixing this for enveloped certs as well.

Thanks!

doomedraven:
```
cert_data['extensions_{}_{}'.format(extension.oid._name, index)] = name.value
```
should be 
```
cert_data['extensions_{}_{}'.format(extension.oid._name, index)] = name.value.rfc4514_string()
```

PR #2984, opened 2020-03-12
Add Localization IP addresses

flavio63730:
Add a localization of the IP addresses that we scan in the network analysis.

To look where does the program want to communicate

It work with different IP addresses.
Does Not display the local network addresses.

### DIFF
--- a/cuckoo/common/files.py
+++ b/cuckoo/common/files.py
@@ -175,6 +175,7 @@ class Files(Storage):
             if not buf:
                 break
             h.update(buf)
+        f.close()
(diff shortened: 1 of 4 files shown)


doomedraven:
you don't need ipinfo at all, just use free maxmind database that can be everything offline and much more faster

botherder:
I agree. I would prefer lookups such as these to be done locally rather than through external APIs. Especially if they are not optional. Additionally, I'd rather see them in the core/modules, rather than the web interface.

Issue #3030, opened 2020-05-13
Netlog client requested banned path: 'package_files/...' for analyzer implementing package_files


omeranson:
Thanks for creating an issue! But first: did you read our community guidelines?
https://cuckoo.sh/docs/introduction/community.html

##### My issue is:

When writing a custom package which implements the `package_files` method, the listed files are not copied. Instead. An error appears in the logs: `Netlog client requested banned path: 'package_files/...'`.

##### My Cuckoo version and operating system are:

Version: 
Cuckoo Sandbox 2.0.7
##### This can be reproduced by:

Using an analyzer implementing `package_files`. 

For instance:

In `cuckoo/data/analyzer/linux/modules/packages/generic.py`, implement the method `package_files`:

```python
# Add this import:
# import tempfile
     def package_files(self):                                                    
         fd, filename = tempfile.mkstemp()                                       
         os.close(fd)                                                            
         return [(filename, 'generic.out')]                                      
```

Expected results: Report will include the file generic.out

Actual results: The file is not included. The following error appears:

2020-05-13 18:30:52,522 [cuckoo.core.resultserver] ERROR: Netlog client requested banned path: 'package_files/generic.out'

PR #3031, opened 2020-05-13
Add package_files path to allowed uploadable paths

omeranson:
In resultserver.py, add 'package_files' to the list of uploadable paths.
This path is used in 'analyzer.py' (both linux and windows versions) when
uploading files reported by the 'package_files' method.

Tested the reproduction instruction in the issue. Reproduced without the change. Got correct results with the change.

Closes: #3030


omeranson:
@botherder Could you please take a look?

Issue #3035, opened 2020-05-20
Error during analysis: could not get process status for pid 


omeranson:
Thanks for creating an issue! But first: did you read our community guidelines?
https://cuckoo.sh/docs/introduction/community.html

##### My issue is:  Error during analysis: could not get process status for pid 

During analysis, the following error appears in the analyser logs: 'Error during analysis: could not get process status for pid ', and the analysis is cut short.

##### This can be reproduced by:

Submitting a file to the generic analyser, where the guest is Ubuntu 18.04.

##### The log, error, files etc can be found at:

The only log is 'CRITICAL: could not get process status for pid XXX' in the analyser log.

Investigation reveals that the error occurs when parsing `/proc/<pid>/status`. One of the fields is empty, and the parsing code assumes that a value always exists. The error happens in the method `get_proc_status`, class `Process`, file: `cuckoo/data/analyzer/linux/lib/api/process.py`.

PR #3036, opened 2020-05-20
linux/process: Support reading empty values from /proc/pid/status

omeranson:
The file /proc/<pid>/status provides information about a process in
the format 'key: value'. The parsing of this file failed if 'value'
was empty.

Add an dummy value that will be taken in case the real value from
the status file does not exist.

Additionally, update the log message from the parsing code to also
output the exception, if one is raised.

Related: #3035

Add an dummy value that will be taken in case the real value from
the status file does not exist.

Additionally, update the log message from the parsing code to also
output the exception, if one is raised.

Maker reading process status data more robust. Have it support cases where some values are empty.

The issue reproduced without the change. Adding the change resolves the issue.


omeranson:
@botherder Could you please take a look?

Issue #3105, opened 2020-09-18
Multiple tasks getting assigned to the same machine


Al33t:
##### My issue is:
When submitting tasks in rapid succession, two tasks can get assigned to the same machine, after which that vm becomes corrupt and needs to be rebuilt. This is usually only observed after a while, i.e. several have to be submitted and then after any length of time this may occurs.

##### My Cuckoo version and operating system are:
Cuckoo version 2.0.7
Debian 10

##### This can be reproduced by:
Submitting several tasks in rapid succession.
One wonders if there is another variable at play since the ease of reproducing this would make it obvious that most users would be affected by this but none of the other reported issues seemed comparable.

The issue has already been analyzed and a race condition was determined to be the cause. By instrumenting the code, a carefully placed delay, the issue got aggrevated to such extent that it was observed sooner and more frequent.

##### The log, error, files etc can be found at:
2020-09-12 03:35:57,259 [cuckoo.core.scheduler] INFO: Task #68871: acquired machine 18 (label=18)
2020-09-12 03:35:57,262 [cuckoo.core.resultserver] DEBUG: Now tracking machine n.n.n.n for task #68871
2020-09-12 03:35:57,263 [cuckoo.core.plugins] DEBUG: Started auxiliary module: Replay
2020-09-12 03:35:57,263 [cuckoo.core.scheduler] INFO: Task #68874: acquired machine 18 (label=18)
2020-09-12 03:35:57,264 [cuckoo.core.resultserver] DEBUG: Now tracking machine n.n.n.n for task #68874
2020-09-12 03:35:57,312 [cuckoo.core.plugins] DEBUG: Started auxiliary module: Replay
2020-09-12 03:35:57,351 [cuckoo.auxiliary.sniffer] INFO: Started sniffer with PID 20509 (interface=vnet18, host=n.n.n.n)
2020-09-12 03:35:57,353 [cuckoo.core.plugins] DEBUG: Started auxiliary module: Sniffer
2020-09-12 03:35:57,389 [cuckoo.auxiliary.sniffer] INFO: Started sniffer with PID 20510 (interface=vnet18, host=n.n.n.n)
2020-09-12 03:35:57,408 [cuckoo.core.plugins] DEBUG: Started auxiliary module: Sniffer
2020-09-12 03:35:57,441 [cuckoo.machinery.qemu] DEBUG: Starting vm 18
[...]

PR #3106, opened 2020-09-18
Fix a race condition on multiple tasks getting assigned to the same machine

Al33t:
This also results in that machine needing rebuilding.

/cuckoo/core/database.py

To avoid a race condition on acquiring machines while they are not yet marked as locked.

Instrumented the code to make the race condition more obvious for testing at runtime.
No specific test case was written for this.

### DIFF
--- a/cuckoo/core/database.py
+++ b/cuckoo/core/database.py
@@ -851,7 +851,7 @@ class Database(object):
                 raise CuckooOperationalError("No machines match selection criteria.")
 
             # Get the first free machine.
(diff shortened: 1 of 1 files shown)


Al33t:
Closes #3105

Al33t:
Closes #3105

Al33t:
Closes #3105

Al33t:
Closes #3105

Al33t:
Closes #3105

PR #3120, opened 2020-10-14
Azure Machinery for Cuckoo

cccs-kevin:

- An Azure machinery for Cuckoo using Azure Virtual Machines
- Ability to dynamically create machines on-demand
- Ability to handle multiple pools of machines that have different OS types
- Ability to specify OS type of VM you want submission to go to

Use cloud resources for malware detonation

Functionality, but only by hand.

#### Things to NOTE:
The Azure API has rate limits to how often you can submit requests. So far the current implementation of this machinery hits the API limit if it process 400 files per hour (each file requires a new VM which is created using a network interface card and a managed disk -> 3 write calls). [Azure doc](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling#subscription-and-tenant-limits)
I'm currently working on a method that utilizes Azure Virtual Machine Scale Sets with Ephemeral disks which avoid hitting this API limit.


Dylan-J:
Man this is brilliant, good work @cccs-kevin!

